Issue #25, opened 2021-04-28
Allow non-Earth based spatial ranges


bodkan:
Not every use-case for this package involves simulating populations on a realistic geographic landscape.

Is it possible to just skip the CRS specification and still do everything we can currently do? Can `sf` handle that?

bodkan:
Closed by #26.

bodkan:
Closed by #26.

PR #26, opened 2021-05-04
Implement abstract "worlds" and landscapes

bodkan:
First attempt at solving #25. Turns out that this needed lots of changes to the internal plumbing, but the good thing is that `sf` can handle all spatial manipulation without any issues. No duplicated handling of "real CRS" and "abstract landscape" situations was necessary. The only difference between these is a presence or an absence of CRS, and whether or not the user specified some kind of landscape or not.


bodkan:
Now let's add some basic unit tests to make sure we catch most of the things that could go wrong when setting up the world context, accidentally including population boundaries with "different worlds", etc....

bodkan:
Closes #25

Issue #5, opened 2021-04-28
Be consistent in the notation of times


bodkan:
It makes sense to specify time in a forwards time notation, if only because this is what SLiM uses in the background. That said, almost all ancient DNA people would prefer to specify times in "years before present", simply because this is how specimens are dated.

Is there a way to have a consistent default (forward time units) but have a switch to reverse this if this is what the user prefers?

Issue #33, opened 2021-05-15
There is something strange going on when times are short/maps duplicated


bodkan:
Look closer at what happens when multiple maps are "mapped" to the same generation on the SLiM side.

PR #38, opened 2021-06-03
Flexible direction of time (forward and backward time convention)

bodkan:


bodkan:
This is an ambitious change in order to close #5.

As part of implementing this, also closing #33.

Issue #61, opened 2021-08-25
save raster information in top-level tree sequence metadata


petrelharp:
It'd be nice to put:
- projection
- raster extent
- ratser resolution
in top-level metadata.  Maybe as
```
"spatial_units" : {
  "type" : "raster",
  "description": <optional name of raster or model or something?>
  "projection" : <string>,
  "extent" : [x0, x1, y0, y1], # or is it [x0, y0, x1, y1] in R?
  "resolution": [nx, ny]
}
```

PR #66, opened 2021-10-06
Refactoring of the SLiM backend and some R internals

### LINKED ISSUE #63: Switch to new msprime.sim_mutations in ts_mutate

bodkan:
This PR started as a way to address @bhaller's comments on the SLiM backend script. I think I addressed and implemented almost every suggested change, most of the points are addressed here one by one. Those which are not I simply implemented without further comments.

This PR also addresses a couple of issues that involved making internal changes to the package (namely issues #63 #62 #61 raised by @petrelharp). Whether these changes have been implemented in a satisfying way could be discussed either in this thread or under respective Github issues themselves.

Feel free to respond or comment to any of these. I don't think there's ton of discussion to be had here, so feel free to also just take this PR thread as a formalized changelog from our last meeting and the emails that followed from that.

There are still some outstanding things in the issue tracker but I wanted to get as much of @bhaller and @petrelharp comments and suggestions implemented before I do anything else. In particular, the SLiM backend is now implemented (and interacted with) in a completely different way, as I describe below.
I'm planning to tackle the remaining couple of issues while I'm writing up the manuscript (doing that as we speak). Same goes for the many comments that @bhaller made on the vignettes.

# Done tasks

---

### Done ✅

> You should add formatted comments for the "jump menu" in SLiMgui.
>
> [...]
>
> You could also annotate the individual script blocks with structured comments
>
> [...]

---

### Done ✅ (and more)

> [...] It'd be nice if your script was syntax-compliant, so maybe you can look for a different way of doing this [`{{var}}`] templating?  One possibility would be to just use a placeholder that already looks like a symbol, like X_X_SEED instead of {{seed}}, or something. [...]

I have now changed the backend from an Eidos/SLiM-noncompliant code (which used those `{{var}}` substitution tags) to normal SLiM command-line parameters. I had a specific idea for the R templating (it was supposed to power those custom user-defined SLiM "modules" I mentioned and make them easy to work with in R) but it's an overkill in this case.

There are three important benefits of his change:

1. When a user `compile()`s a slendr model, a dedicated SLiM script will be saved in the model directory together with all other config files and can be run from the command-line as any other SLiM script. The `slim()` runner function in slendr now simply passes all the parameters to the SLiM backend script via command-line instead of substituting/hardcoding the values via the template `{{var}}` variables (as in the previous version). The main benefit is a significant increase in flexibility. A complete model can now be programmed and compiled on one machine (in R), and then transfered to another machine for execution. Some institutions have very restricted cluster capabilities--this is one way to get around that, because those environments will now only require a compiled SLiM binary, nothing else.

2. The `compile()` function now accepts an optional argument specifying where it should look for the backend SLiM script. By default, it's the script bundled under `inst/extdata/script.slim` in the R package. But it could be also a slightly modified version of this script, with bits of added functionality (custom script blocks, etc.). As long as the core of the script remains the same, any advanced SLiM user can extend the functionality however they need, and still maintain the reproducibility of having the entire analysis/simulation workflow done from R (the only thing that changes is specifying `script_path = <...>` in the `compile()` function).

3. This solves my issue with those optional "modules". Before this change, the script was always saved as one "monolithic" unit with all parameter values hardcoded via the `{{var}}` substitution by the `slim()` R function "at runtime". The users still had the option to modify the compiled scripts, but it would break the reproducibility of the whole workflow (those changes would happen outside of slendr). This is not the case anymore. The users can add modifications to the SLiM script, let slendr know that this custom script should be `compile()`-d together with the model data, and the `slim()` runner function will simply use it, keeping the whole reproducible R pipeline intact.

---

### Done ✅

> You seem to be using two-space indents, I'd suggest tabs?

I hope this won't ruin our friendship, but I'm a card-carrying member of the spaces-but-not-tabs camp. 😬 I do admit that the two-space indents come from an R style guide which doesn't fit SLiM code (or non-R code in general). I changed the code to four-space indents. Hopefully a reasonable compromise.

---

### Done ✅

> You do "N = event.getValue("N");", so N is a variable. [...] how about "pop_size" or some such?

OK, I made this change. However, given that UPPER_CASE_CONSTANTS are now basically all command-line arguments, I'm actually thinking about making them lower-case. I personally don't see much of a reason to call the default slendr SLiM script outside of R for 99% of use-cases, but having to type out `-d UPPER_CASE_CLI_ARGUMENT=123` somehow bothers me a bit. A matter of esthetics, I guess.

---

### Done ✅

> add_markers() checks if (TRACK_ANCESTRY) and does nothing if it's false.  Seems weird to me; more normal would be to put the if (TRACK_ANCESTRY) outside the function, in the caller not the function.  This is because the function name suggests that the function simply "adds markers" [...] A nit, but this sort of thing can cause confusion.  Functions/methods should do what their name says they do, to the extent possible.  Sorry for being pedantic, I feel a little guilty about this comment.  :->

No reason to feel guilty. As a fellow code pedant, I appreciate this comment. 👍 In my semi-defense, I have no idea why this condition was even there. I think it must be a left-over of a pre-table/Dictionary iteration of the backend which probably did something slightly more in this function.

---

### Done ✅

@peterlharp Does this look like a reasonable way to do #61 and #62?

We have a `collect_metadata()` function [on the backend](https://github.com/bodkan/slendr/blob/backend-refactoring/inst/extdata/script.slim#L580) which generated metadata to be saved [in the tree sequence output](https://github.com/bodkan/slendr/blob/backend-refactoring/inst/extdata/script.slim#L424). The metadata placeholders (as well as the slendr version string) are substituted in the backend script [during compilation](https://github.com/bodkan/slendr/blob/backend-refactoring/R/compilation.R#L488).

I modified the format of the metadata slightly (moved `"description"` one level "higher" in the hierarchy directly under the `"slendr"` key) and simplified it only to what's currently used by slendr as of now. It will be quite easy to expand/modify the format should other spatial tools pop up at some point.

---

### Done ✅

Mutations are now overlaid using `msprime.sim_mutations` (closes #63).

---

### Done ✅

> slim_time(), yikes.  That's a lot of complicated code just to translate times.  Would it be possible to do a different approach to this?  You could maybe (a) write out the translated times in the first place so they don't need to be shifted at runtime, or (b) do the shift by simply adding BURNIN_LENGTH when you *use* a time value that you just got from a table?  Don't know if any of this is practical, but this sort of complex dictionary-munging is definitely an obstacle to understanding how the script works.

~~Three paragraphs of excuses and justification for why I can't or don't want to address this right now.~~

This was an excellent point. It took me a whole week sitting on this, while I was implementing your other comments but then I found a good way to implement your suggestions (option (b)). This made the code *so much* cleaner. Thank you!

---

### Done ✅

> original_time(), I'm not really sure what it does.  In what units is "current generation time", in what units is "the original units", why are they different, etc.?  The comment on this function needs to be improved.

Same as the above. It was surprisingly painful to do this, but I'm so glad I did it. Lots of code was removed by doing this -- the best kind of refactoring. :)

### Done ✅

> In your modifyChild() callback you do "return F;" if a location is generated that can't be used. [... ]I'd suggest "reprising boundaries", which would loop until a legal position is generated.  If it is plausible that that could be an infinite loop, then you could set a maximum number of tries by using a for loop with a return on success, and return F if the for loop completes.

That's a good point, I implemented reprising boundaries exactly as you suggested, making the "number of attempts before failing" an optional parameter I have no intuition as to what is and what isn't reasonable default here, so I put it at 10 and gave the user the option to change this if needed.

---

# Tasks in progress

---

### In-progress ⏳

> I wonder whether any of the output you do should be done with LogFile?  Or if you might want to add the option for additional output of runtime stats/metrics using LogFile?

That's something Peter also suggested at some point but I have never managed to get to this. Now that we are close to the final feature set, this is definitely a quality of life improvement worth revisiting. I will read the appropriate section of the manual and see what I can do (in a separate PR, probably).

---

### In-progress ⏳

> set_coordinates() could be made vastly more efficient by vectorizing it, but if it's not a bottleneck don't worry about it, because the way to vectorize it is a little bit gross (see https://github.com/MesserLab/SLiM-Extras/blob/master/models/Recipe_15.10_OnLand.slim)

Ah, this is clever. Also, yet another case where I missed a clear vectorization possibility. I actually think it makes sense to optimize the code as much as possible. I will do some profiling (I've never actually used the SLiM profiler before 😬) first and will tackle this in a separate PR. I might also just wait and see if there are problems down the line, as this would be simply an internal change, not something affecting potential slendr models in the wild.

---

> Note that "asInteger({{seq_length}})" is unnecessary unless the string you put in place of {{seq_length}} is not an integer (which seems like it would be a bug?).  In Eidos, unlike in R, 10 is an integer, not numeric, and in fact 5e5 is also an integer.  Ah, but 5.5e5 is a float, which is arguably a bug in Eidos; that should be an integer too.  So OK, maybe you need that call.  Just thinking out loud, and now I'm curious whether you guys agree that 5.5e5 being float is a bug in Eidos.  :->

My vote would be for 5e5 being an integer and not a float, if only because the "an integer-looking number is numeric but not integer by default" thing in R always bothers me a bit.

---

### Zoom call? ☎️

> I see that you do "initializeMutationRate(0.0);".  So slendr simulations never have new mutations at all?  It looks like you define genomic element types only for ancestry inference, when tree sequence recording is turned off I suppose?  And I guess neutral mutations can be overlaid, with treeseq at least.

In my original plan for slendr, mutation rate was supposed to be a flexible parameter. I was aiming to have those user-defined SLiM "modules" for selection etc., which would include user-defined non-neutral mutation types. Now that this is off the table for the nearest future, I'm not sure anymore.

I've become such a big fan of overlaying mutations after the simulation that I have fully embraced tree sequences as the default output of slendr models, implicitly assuming that neutral mutations will be overlaid later. On the other hand, there's no reason to have one mutation type for accumulating mutations during the SLiM run, tied to an optional mutation rate parameter, so I just added that. Do we actually need this? I personally don't, but perhaps someone else will. Maybe there should then be an option to save a VCF with the mutations accumulated through a SLiM run, too (sidestepping the tree sequence processing completely)? In case someone is not interested in tree sequence things. That said, I think the tskit interface in slendr is so easy that it doesn't add any complexity for a user other than requiring them to run `pip install tskit pyslim`.

Thoughts? @petrelharp @bhaller @ferracimo

Also, I'm not sure what to do with the ancestry-tracking neutral markers. I honestly don't want this in the backend but perhaps I should leave things as they are until there's a more clever way to do this with tskit.

---

### Zoom call ☎️

> I don't think I'm going to try to thoroughly grok your whole script; it's too complex and difficult to understand since it is table-driven etc.  That's nothing against your code or design, it looks very clean to me.  It just takes a lot of work to get deeply into a script like this without the help of the author.  :->  I'd be happy to have a zoom with you where we go over the script line by line, though; that would probably be useful for both of us.  Anyway, I'm going to limit myself to more superficial comments based on what I see just skimming your code without trying to deeply understand it.

I agree, a separate Zoom call where we could go through this would be great.

---

### Zoom call ☎️

> Your 2: fitness(NULL) callback would probably be considerably more efficient if it were rewritten to (a) not be a callback, (b) use fitnessScaling, and (c) be vectorized.  I don't know whether speed is an issue for your stuff or not.  If it is, you should do a profile in SLiMgui to see where your bottlenecks are, and then perhaps I can help you whittle them down.

I will look at this soon. 👍

---

### Zoom call ☎️

> calc_ancestry(), interesting.  I really wish we could do this with tree-sequence recording.

Me too! I noticed you discussed this with Peter in our email thread recently. This would be an *amazing* feature to have and--as I mentioned above--I would actually prefer not to have to do the neutral ancestry markers shenanigans in the backend script. I imagine it would speed things up quite a bit too.

---

### Zoom call ☎️ (?)

> Your functions to work with Dictionaries are interesting.  read_table() could instead use Dictionary's JSON-string constructor, if you wrote out your tables as JSON rather than .tsv.  But maybe being able to read a Dictionary from a .tsv is also useful functionality that ought to be in Eidos.  Not sure.  So far my focus has been on JSON because that's what Graham Gower seems to be interested in using.

Good point on reading the tables as JSON. AFAIK this package started a bit before the JSON-loading functionality landed in Eidos and I never made the switch from the plain text tables, mostly because sticking to tables made it very easy to debug problems, even for complex models.

My mid-to-long-term plan is to save slendr config data in a Demes-like format--Demes for the non-spatial part, and some slendr-specific extension for the spatial bits. But for now I'd leave things the way they are and only switched the backend config format once things settle down for the first release.

--- 

> - filter(), also interesting.  Something like this could also conceivably be added to Eidos.
> 
> - num_rows(), ditto.  Possibly this idea of a Dictionary as a dataframe-like object ought to be more formalized.  Eidos could define a subclass of Dictionary named DataFrame that defined more of this sort of functionality.
> 
> - print_table(), ah, interesting.  Yes, the default print for Dictionary is not so readable when the Dictionary is, conceptually, a dataframe.  Again, if DataFrame were a subclass this could be handled in Eidos.

These would be very useful additions indeed, even after I make the JSON switch once the slendr "format" settles down.

---

### Zoom call ☎️

> `convert_type()`, eww.  :->  I'd like to better understand exactly why this is needed.  It might be that if you switched to JSON-serialized tables the need for this would disappear.

I'm not happy with this piece of code either.

Why this is currently needed: unless I completely missed something, the `read_table` function doesn't have a way to detect which columns are of which type. It uses `readFile()` to read the whole file, converting it to a matrix via `strsplit()` call. But my attempts at figuring out which type (integer, float, string, logical) is particular column of using Eidos code at runtime have failed (maybe I missed something).

Given that the long-term plan is saving the model config as JSON files (unless some .tsv support lands in Eidos before that), I'll probably leave this as it is.


bhaller:
@bodkan all sounds good.  I have opened https://github.com/MesserLab/SLiM/issues/230 on the 5.5e5 issue.

petrelharp:
> I'm actually not even sure how useful this real-time tracking would be.

Agree - if you don't actually have a use case for it now, then definitely don't do it.

Issue #85, opened 2022-03-28
Generalize the R-tskit interface towards non-slendr tree sequences


bodkan:
Now that we can load and analyse both SLiM and msprime tree sequences originating from slendr models, it's time to generalize the [`ts_*()` interface](https://www.slendr.net/reference/index.html#tree-sequence-loading-and-processing) so that the users can analyse even non-slendr tree sequences produced by standard non-slendr msprime and SLiM simulations.

This will get us towards slendr that's not just an R-friendly interface for configuring and simulation a specific subset of demographic models in msprime and SLiM, but also a convenient R environment for loading/processing/analyzing/visualizing tskit tree sequences in general.

Especially for spatial models, the benefits having `ts_data()`, `ts_tree()`, `ts_ancestors()`, etc. return proper _sf_ data tables which can be analyzed with the myriad of R packages for geospatial statistics and phylogenetics, would be numerous. People do lots of amazing things with spatial SLiM, for instance, things which slendr is unlikely to support.

---

Part of this effort is to have this generalized interface also compatible with tsinfer and tsdate outputs. Hopefully these are just any other tskit tree sequences. One challenge here will be correctly translating metadata (for now, lat/lon locations and times of nodes). But that should be a solvable problem that I will tackle in a separate issue.

bhaller:
I wonder whether it might make sense to coordinate with @petrelharp to try to make slendr's front end to tree sequences be as similar to pyslim's front end as possible?  Is that a worthwhile goal, or are these really just two completely orthogonal approaches/uses/functionalities?

bodkan:
Hmm, good point. Certainly, there shouldn't be any case where the slendr tskit-R interface would be doing something that would surprise the user who would be familiar with tskit and/or pyslim Python APIs. This is why I have tried to use the same default argument values in all current tskit-R functions of slendr (where applicable).

That said, the interface we have [right now](https://www.slendr.net/reference/index.html#tree-sequence-loading-and-processing) is heavily designed to be R-like. If only because R doesn't really support the kind of OOP style that Python prefers and the way R is often used (being a functional programming language as opposed to primarily OO language) is quite different from a Python approach to data analysis.

[...]

PR #91, opened 2022-04-28
Generalize the R-tskit interface towards non-slendr tree sequences

bodkan:
Finally some solid progress towards #85.

We can now:

- load a standard (non-slendr) SLiM and msprime tree sequence (`ts_load()`)
- simplify a non-slendr tree sequence (`ts_simplify()`)
- convert one tree from the tree sequence to a standard R phylogenetic ape format phylo (`ts_phylo()`)

Not completely happy with the code. Basically, I'm currently adding things like

```
if (!is.null(attr(ts, "model"))) { // if the tree sequence does not have slendr-specific attribute...
    // ... do some slendr-specific processing
}
```

Which is not pretty because these are often scattered throughout those functions.

The "specific processing" above generally involves:

- converting tskit node ages to "slendr-model-specific" time units
- sprinkling slendr's nice readable symbolic names for individuals on top of the raw tree-sequence table data.

Next steps:

- Make sure `ts_recapitate` does what it's supposed to do even on non-slendr tree sequences.
- Implement conversion of SLiM tree-sequence individual coordinates (raster-based) to the _sf_ format. This will enable SLiM users who don't use slendr for simulations access all the awesome geospatial libraries for R via the slendr R-tskit interface. They will simply get an _sf_-compatible data frame with times and locations of everything recorded in their tree sequence.
- Implement a variation of tree-sequence unit tests already in slendr. These will use simple SLiM scripts and `msprime.sim_ancestry(N)` calls to generate SLiM and msprime data and make sure that `ts_data()`, `ts_samples()`, and friends work as they should even on those.

I already tested the tskit popgen statistics methods (`ts_divergence()`, `ts_f4()`, etc.) and they seem to work without me having to make any changes. This was a pleasant surprise: I had forgotten that those functions already worked on non-symbolic integer tskit node indices, so we should be good on that front. Good for me. :)


bodkan:
Hello @bhaller, I have quite an exciting update to share. Also tagging @FerRacimo so he is kept in the loop, because we spoke about this a couple of times in the past.

I have now finished a draft implementation of the tskit-powered tree-sequence functionality in slendr towards _non-slendr_ tree sequences. I.e, it is now possible to take any non-slendr tree sequence (either from SLiM or msprime), load/simplify/recapitate/process it with slendr in R, and run tree-sequence popgen statistics on it (at least those currently supported by slendr). Spatial tree sequences are also supported, so all the geospatial R goodies are now available.

This gets us significantly closer towards having slendr as a toolkit for analyzing tree sequence data in general (non-spatial and spatial!). In fact, barring any bugs (still working on unit tests), there should be anything in the tskit-R interface of slendr that would work on slendr tree sequences at the moment but wouldn't work on non-slendr tree sequence.

### A couple of quick examples:

#### Non-spatial SLiM (non-slendr) tree sequences

Consider the following simple SLiM script, which creates a couple of subpopulations (with different Ne) splitting from an ancestral p1 subpopulation:

```
initialize() {
	initializeTreeSeq();
	initializeMutationRate(0);
	initializeMutationType("m1", 0.5, "f", 0.0);
	initializeGenomicElementType("g1", m1, 1.0);
	initializeGenomicElement(g1, 0, 1e6);
	initializeRecombinationRate(1e-8);
}

1 {
	sim.addSubpop("p1", 10);
}

1000 {
	sim.addSubpopSplit("p2", 500, p1);
}

3000 {
	sim.addSubpopSplit("p3", 2500, p1);
}

5000 {
	sim.addSubpopSplit("p4", 10000, p1);
}

15000 late() {
	sim.treeSeqOutput("~/Desktop/nonspatial.trees");
}
```

We can use slendr to load the output tree sequence, simplify it, and overlay mutations on it using the standard functionality originally developed for to slendr tree sequences:

```
ts <- ts_load("~/Desktop/nonspatial.trees") %>%
  ts_simplify() %>%
  ts_mutate(mutation_rate = 1e-7)
```

We can extract information about individual's names, nodes, population assignments, etc. just as with any slendr tree sequence (`ts_data()` basically loads raw nodes and individuals tree sequences tables, performs a couple of join operations, and presents the whole thing in a nice unified form for interactive data analysis):

```
data <- ts_data(ts) %>% dplyr::filter(sampled)
data

> slendr 'table' object 
> --------------------- 
> data was extracted from a SLiM forward time model
> 
> summary of the table data contents:
> 
> total:
>   - 13010 'sampled' individuals
>   - 0 'retained' individuals
>   - 0 no nodes from 'recapitated' individuals
> --------------------- 
> oldest sampled individual: 0 time units 'before present' 
> youngest sampled individual: 0 time units 'before present' 
> 
> oldest node: 0 time units 'before present' 
> youngest node: 0 time units 'before present' 
> --------------------- 
> overview of the underlying table object:
> 
> # A tibble: 26,020 × 9
>    pop   ind_id node_id  time sampled remembered retained alive pedigree_id
>    <chr>  <dbl>   <int> <dbl> <lgl>   <lgl>      <lgl>    <lgl>       <dbl>
>  1 p1         0   20000     0 TRUE    FALSE      FALSE    TRUE    137163000
>  2 p1         0   20001     0 TRUE    FALSE      FALSE    TRUE    137163000
>  3 p1         1   20002     0 TRUE    FALSE      FALSE    TRUE    137163001
>  4 p1         1   20003     0 TRUE    FALSE      FALSE    TRUE    137163001
>  5 p1         2   20004     0 TRUE    FALSE      FALSE    TRUE    137163002
>  6 p1         2   20005     0 TRUE    FALSE      FALSE    TRUE    137163002
>  7 p1         3   20006     0 TRUE    FALSE      FALSE    TRUE    137163003
>  8 p1         3   20007     0 TRUE    FALSE      FALSE    TRUE    137163003
>  9 p1         4   20008     0 TRUE    FALSE      FALSE    TRUE    137163004
> 10 p1         4   20009     0 TRUE    FALSE      FALSE    TRUE    137163004
> # … with 26,010 more rows
```

Moving on to tskit statistics, we can use the data table above to extract a list of nodes belonging to each population (this is what various tskit tree-sequence statistics operate on, and slendr follows that design). Here we are computing the nucleotide diversity in each of the four populations using the `ts_diversity()` function:

```
sample_sets <- split(data$node_id, data$pop)

# compute nucleotide diversity in each population
# (any other ts_*() tskit R interface function should work)
ts_diversity(ts, sample_sets)

> # A tibble: 4 × 2
>   set    diversity
>   <chr>      <dbl>
> 1 p1    0.00000371
> 2 p2    0.000192  
> 3 p3    0.000900  
> 4 p4    0.00161   
```

Just as with slendr tree sequences (as demonstrated in our preprint) we can get a individual trees too, extracted in the in the phylogenetic format provided by the ape R package. Here we first simplify the tree sequence even further to just 10 nodes to make things manageable:

```
samples <- sample(data$node_id, 10)
ts_small <- ts_simplify(ts, simplify_to = samples)

# extract the 42nd tree in the genealogy to an R 'phylo' format
tree <- ts_phylo(ts_small, 42)
tree

> Phylogenetic tree with 10 tips and 9 internal nodes.
> 
> Tip labels:
>   2, 6, 3, 9, 5, 4, ...
> Node labels:
>   209, 13, 17, 28, 129, 200, ...
> 
> Rooted; includes branch lengths.
```

Once we have that R tree object, we can use packages like ggtree to visualize the tree (any other phylogenetic package would work too). Note that because nodes of 'ape phylo' trees must conform to a strict format (they must be labelled 1...N), we will extract the information about the node IDs in the tskit tree sequence data to be able to plot them in the tree.

Significantly less eye candy than the tree in our preprint, but it conveys the idea:

```
labels <- ts_data(tree) %>% select(node = phylo_id, tskit_id = node_id)

ggtree(tree) %<+% labels +
  geom_label(aes(label = tskit_id))
```

image

#### msprime (non-slendr) tree sequences

The whole thing works also for msprime tree sequences (not super surprising, given that it's all tskit under the hood).

---

#### Spatial SLiM (non-slendr) tree sequences
Furthermore, the generalized interface also supports slendr's spatial tree-sequence features, with all bells and whistles.

If we take the following spatial model (modified from the SLiM manual):

```
initialize() {
	initializeSLiMOptions(keepPedigrees=T, dimensionality="xy");
	initializeTreeSeq();
	initializeMutationRate(1e-7);
	initializeMutationType("m1", 0.5, "f", 0.0);
	initializeGenomicElementType("g1", m1, 1.0);
	initializeGenomicElement(g1, 0, 1e6);
	initializeRecombinationRate(1e-8);
}
1 early() {
	sim.addSubpop("p1", 500);
	
	// initial positions are random in ([0,1], [0,1])
	p1.individuals.x = runif(p1.individualCount);
	p1.individuals.y = runif(p1.individualCount);
}
modifyChild() {
	// draw a child position near the first parent, within bounds
	do child.x = parent1.x + rnorm(1, 0, 0.02);
	while ((child.x < 0.0) | (child.x > 1.0));
	
	do child.y = parent1.y + rnorm(1, 0, 0.02);
	while ((child.y < 0.0) | (child.y > 1.0));
	
	return T;
}
1: late() {
    sim.treeSeqRememberIndividuals(sim.subpopulations.individuals, permanent = F);
}

10000 late() {
	sim.treeSeqOutput("~/Desktop/spatial.trees");
}
```

We can load and simplify the output tree sequence in the standard slendr way:

```
ts <- ts_load("~/Desktop/spatial.trees") %>% ts_simplify()
```

We can then access the spatio-temporal data embedded in the output tree sequence in the standard slendr way (note the "spatial" sf column `location` with the `POINT` data type):

```
data <- ts_data(ts)
data

> slendr 'table' object 
> --------------------- 
> data was extracted from a SLiM forward time model
> 
> summary of the table data contents:
> 
> total:
>   - 500 'sampled' individuals
>   - 777 'retained' individuals
>   - 0 no nodes from 'recapitated' individuals
> --------------------- 
> oldest sampled individual: 0 time units 'before present' 
> youngest sampled individual: 0 time units 'before present' 
> 
> oldest node: 1355 time units 'before present' 
> youngest node: 0 time units 'before present' 
> --------------------- 
> overview of the underlying sf object:
> 
> # A tibble: 1,817 × 10
>    pop   ind_id node_id  time               location sampled remembered retained alive pedigree_id
>    <chr>  <dbl>   <int> <dbl>                <POINT> <lgl>   <lgl>      <lgl>    <lgl>       <dbl>
>  1 p1         0       0     0  (0.3744914 0.2729115) TRUE    FALSE      TRUE     TRUE      5000000
>  2 p1         0       1     0  (0.3744914 0.2729115) TRUE    FALSE      TRUE     TRUE      5000000
>  3 p1         1       2     0 (0.2956575 0.03230727) TRUE    FALSE      TRUE     TRUE      5000001
>  4 p1         1       3     0 (0.2956575 0.03230727) TRUE    FALSE      TRUE     TRUE      5000001
>  5 p1         2       4     0 (0.3947814 0.02905613) TRUE    FALSE      TRUE     TRUE      5000002
>  6 p1         2       5     0 (0.3947814 0.02905613) TRUE    FALSE      TRUE     TRUE      5000002
>  7 p1         3       6     0  (0.1304009 0.1380062) TRUE    FALSE      TRUE     TRUE      5000003
>  8 p1         3       7     0  (0.1304009 0.1380062) TRUE    FALSE      TRUE     TRUE      5000003
>  9 p1         4       8     0  (0.4263232 0.3269192) TRUE    FALSE      TRUE     TRUE      5000004
> 10 p1         4       9     0  (0.4263232 0.3269192) TRUE    FALSE      TRUE     TRUE      5000004
> # … with 1,807 more rows
```

Because we get the tree sequence converted to the spatial _sf_ data format, we can use standard geospatial packages to use any spatial data analysis methods that those packages provide.

Just to demonstrate, we can trivially plot the location of each recorded node (ggplot is not mandatory, it's just something I'm familiar with!):

```
ggplot() + geom_sf(data = data, aes(color = time), alpha = 0.5)
```

image

We can also collect spatio-temporal ancestry information of a particular node (i.e. the times and locations of all of its ancestors all the way to the root, with each "link" in the plot signifying parent-child edge somewhere along the tree sequence) and plot it on a 2D surface (x and y dimensions [0, 1]). The plot is obviously chaotic, but should convey the idea (the "focal node" 0 is highlighted in red). It's the same plot we have in the last figure of our paper.

```
ancestral_links <- ts_ancestors(ts, 0)

ggplot() +
    geom_sf(data = ancestral_links, size = 0.5, aes(alpha = parent_time)) +
    geom_sf(data = sf::st_set_geometry(ancestral_links, "parent_location"), aes(color = parent_time)) +
    geom_sf(data = data[data$node_id == 0, ], size = 3, color = "red")
```

image

bodkan:
I will probably collect those simple examples in the previous message into a separate vignette. It won't add much that wouldn't be covered by the other vignettes, but it might be helpful to the people who use msprime or SLiM _without_ slendr (basically everyone at this point) that there is a possibility to simulate data with whatever means and still analyze them in R via slendr's tskit interface.

bodkan:
Ugh. Messed up tagging (wrote this in RStudio and only then copied over to GitHub).

Hello @petrelharp and @mkiravn, I have started pushing updates that you both might find relevant and interesting. [Here](https://github.com/bodkan/slendr/pull/91#issuecomment-1114329598) is a link to a short demonstration. We seem to be able to finally load/recapitate/simplify non-slendr tree sequences from "pure" SLiM and msprime simulations and analyze them with our tskit R interface. Still more work to do to make sure everything works as it should but we're getting there!

bodkan:
Also tagging @gregorgorjanc to [check out the examples](https://github.com/bodkan/slendr/pull/91#issuecomment-1114329598). Just so that he knows that things are slooowly moving forward towards generalizing our R-tskit interface as promised. :) 

Hoping to merge this PR soon so that other people can test it and break it. I don't use "pure" msprime and SLiM that often, so I'm fairly sure there are things I've missed.

bhaller:
This all looks amazing @bodkan!  As I mentioned in an email to you just now, it would be great to have a recipe in the SLiM manual that shows off the power of this for analysis of SLiM models.  Let me know how you'd like to proceed – I could basically just turn what you wrote above into a recipe, if you think that would be good.  Anyhow, congratulations, this is a big step forward!

bodkan:
> This all looks amazing @bodkan! As I mentioned in an email to you just now, it would be great to have a recipe in the SLiM manual that shows off the power of this for analysis of SLiM models. Let me know how you'd like to proceed – I could basically just turn what you wrote above into a recipe, if you think that would be good. Anyhow, congratulations, this is a big step forward!

Thank you, that's a splendid idea! I think it would be great to turn these examples into a recipe in the SLiM manual. I think they show rather nicely how this whole thing works and what can be expected from it and -- most importantly -- it hopefully doesn't require _too much_ code for a SLiM user to find overwhelming. In my experience, this is always the biggest issue whenever anything "geospatial" is mentioned anywhere. This stuff is not hard but it does often require quite a bit of code and prior knowledge. Fortunately, lots of this hard work is done by slendr internally.

I expect a couple more days of work on this. Then I will update those examples to reflect the new API and let you know. Nothing dramatic, I just noticed there are some redundancies and inconsistencies here and there that should be cleaned up (`ts_edges` vs `ts_branches` distinction should not be necessary and, in general, the table access from R can be simplified quite a bit compared to what it is now).

I think it also makes sense to turn these examples into a tiny vignette so we can work on this together afterwards. You could then simply include the final outcome of that in the manual. Let's chat about the vignette and the SLiM recipe in the email thread we started on this once things settle down a bit in this PR.

petrelharp:
This looks **very** nice!! I'll have a look through the code.

> Not completely happy with the code. Basically, I'm currently adding things like
```
if (!is.null(attr(ts, "model"))) { // if the tree sequence does not have slendr-specific attribute...
    // ... do some slendr-specific processing
}
```
> Which is not pretty because these are often scattered throughout those functions.

You've probably thought about making slendr-produced tree sequences a subclass of the more general tree sequence? (maybe even TreeSequence > SLiMTreeSequence > SlendrTreeSequence) It would probably be more trouble than it's worth, though - in pyslim I started out that way, but now need to un-do that, because it's caused lots of trouble.

R/tree-sequences.R
```diff
@@ -83,30 +111,20 @@ ts_load <- function(model, file = NULL,
...
+  # is good enough for now
+  provenance <- ts$provenance(0L)$record
+  if (grepl("SLiM", provenance)) {
     ts <- pyslim$SlimTreeSequence(ts)
     backend <- "SLiM"
-  } else
+  } else if (grepl("msprime", provenance))
     backend <- "msprime"
+  else
+    stop("Only tree sequences produced by SLiM or msprime are supported", call. = FALSE)
```
petrelharp, line 127: The right way to check if something is SLiM-produced is to check in top-level metadata? (i.e, check for `ts.metadata['SLiM']`) However, do you really need to check this at all? You don't want to stop someone else from using slendr on a valid tree sequence made some other way?

R/tree-sequences.R
```diff
@@ -120,26 +138,31 @@ ts_load <- function(model, file = NULL,
...
-    attr(ts, "data") <- get_slim_table_data(ts, model, spatial)
+    remembered <- attr(ts, "raw_individuals")$remembered
+    if (!any(remembered)) remembered <- attr(ts, "raw_individuals")$alive
+    attr(ts, "raw_individuals")$sampled <- remembered
+    attr(ts, "nodes") <- get_slim_table_data(ts, model, spatial)
   } else {
-    attr(ts, "individuals") <- get_ts_individuals(ts)
-    attr(ts, "data") <- get_msprime_table_data(ts, model)
-    }
+    attr(ts, "raw_individuals")$sampled <- TRUE
```
petrelharp, line 163: Isn't the right way to do this to use the individuals' node flags?

R/tree-sequences.R
```diff
@@ -323,27 +348,25 @@ ts_simplify <- function(ts, simplify_to = NULL, spatial = TRUE, keep_input_roots
     return(ts)
   }
 
-  if (is.null(simplify_to) && backend == "SLiM")
+  if (is.null(simplify_to)) {
     samples <- dplyr::filter(data, sampled)$node_id
-  else if (is.character(simplify_to)) {
+  } else if (is.character(simplify_to)) {
+    if (is.null(model))
```
petrelharp, line 354: seems like you need a better way to check "is this slendr-generated" than `!is.null(model)`

R/tree-sequences.R
bodkan: Hmm, I don't think `ts.metadata["msprime"]` is a thing though? At least in my msprime runs that field is empty? I thought that "provenance" is exactly what should specify the source of the tree sequence (and checking the dictionary to see what the word "provenance" even means confirmed my assumption :D). If whatever produced a tree sequence doesn't populate the metadata field (as seems to be the case for msprime), then I can't know the source right?

Or is it that there are only two classes of tree sequences out there? SLiM and everything else? In that case, it would be enough to have SLiM-specific table munging code and the msprime-specific code could be used for msprime and everything else. Rather than assuming "everything else" will require dedicated code (on top of SLiM-specific and msprime-specific code).

I guess I don't know quite enough about all other possible sources of tree sequence data out there.

R/tree-sequences.R
bodkan: Yes, the individuals' node flags are always present and accessible to the user. However, I discovered that a tree sequence in which every (or most) individuals are "remembered" breaks lots of my code.

For instance, attempting to extract a single tree from such tree sequence and converting it to a standard phylogenetic tree object (all that I evaluated assume only tips are samples) broke because even if a user specified only a subset of "remembered" individuals to be used for simplification (out of every other node which was also "remembered"), other internal nodes in the simplified tree sequence were also remembered... and slendr had no way of knowing which of them are those that the user is interested in (i.e. the subset they provided to guide the simplification process) and which are "not interesting" -- internal nodes NOT provided for simplification -- but still carrying the remembered flag (because every single individual has been remembered).

This is why I added a separate tag that I'm carrying along in the slendr-munged tables called `sampled`. When a tree sequence is first loaded, every "remembered" individual assumed to be "sampled" (because the only thing this can happen in slendr is by being marked as sampled from the R interface), but when that tree sequence is simplified down to a subset of (potentially very many) remembered nodes, slendr keeps track of which nodes are interesting to the user (called "samples") and which are not (everything that's not `sampled == TRUE`).

R/tree-sequences.R
bodkan: I'm utterly confused by GitHub's PR commenting interface (first time using it!) so I'm just acknowledging you pointing this out. I will check what this looks like in the current version of the code and fix things as needed. Thanks!

R/tree-sequences.R
bodkan: I do admit that it's been all motivated by being able to use these "edge-case" tree sequences (i.e. everyone who ever lived is explicitly remembered in the tree sequence) with the standard phylogenetics packages I've looked at.

Maybe there's a better solution that I'm not seeing right now. But

a) all the pyslim flags are still present and available (I'm not reinventing them here)
b) this is only needed in a couple of places in the whole codebase, so if a better way to do this comes up it's not like the whole thing will have to be re-written from scratch (the "sampled" column became a thing only 2 months ago after  months of not being needed... until I hit that "everyone is remembered" fringe case).

bodkan:
> You've probably thought about making slendr-produced tree sequences a subclass of the more general tree sequence? (maybe even TreeSequence > SLiMTreeSequence > SlendrTreeSequence) It would probably be more trouble than it's worth, though - in pyslim I started out that way, but now need to un-do that, because it's caused lots of trouble.

Yeah, that thought did cross my mind but I think I've managed to avoid doing that. In a sense, the "object is a list with a special class attribute" model of S3 OOP in R worked very nicely here, without formalizing the OOP hierarchy like what one would do in Python.

I.e., in R I already set `class` of each slendr-loaded tree sequence to be `"slendr_ts"` on top of the usual `"tskit.trees.TreeSequence"` which is passed through Python via reticulate, which I guess one could say is an equivalent of subclassing in other OOP systems? Maybe I'm wrong, it's been a couple of years since I subclassed anything (last time in C++). :)

Plus, the extra things I need are tagged along under `attr(<ts_object>, "attribute_name")` (such as the model object which generated the tree sequence, what was the sampling schedule used in the simulation, that sort of thing). Probably a horrid thing from the perspective of more formalized and rigorous OOP systems but... it works. ¯\\_(ツ)\_/¯

R/tree-sequences.R
petrelharp: Here's the situation: msprime tree sequences have just the basic stuff; SLiM tree sequences have extra metadata of a particular sort. There may be other sort out there, but none that slendr has to worry about, that I know of. So, I think you can have (a) code that applies to all tree sequences, and (b) code that does extra stuff for SLiM tree sequences.

But - what's the 'extra stuff' you do for SLiM?

R/tree-sequences.R
petrelharp: Gee, this is confusing. So, what's the problem with 'everyone remembered' tree sequences? What breaks when you try to get a tree?

R/tree-sequences.R
petrelharp: I've just noticed that you're doing `is.null(model)` a lot, and I think it's to mean "is this slendr-generated"; it seems like a good thing to abstract out?

R/tree-sequences.R
bodkan: This is very helpful, thank you!

If msprime tree sequences are no different from any other tskit tree sequence (either constructed manually, or obtained through something like tsinfer, etc.) then yes, anything that has something to do with msprime can be used for "generic" tree sequences, no problem there.

Main extra stuff that happens for SLiM tree sequences is:

- If those are generated from slendr models, that "extra stuff" would be:
    - Assigning symbolic names to individuals (and nodes) so that they can be referred to by users in simplification or computing statistics, outputting VCF or EIGENSTRAT, directly by names (i.e. `EUR_42` based on a user-defined sampling table, instead of numerical node IDs).
        - Also maintaining those names through simplification (because tskit renames node/individual IDS, I'm using `pedigree_id` from SLiM through join operations to know what got dropped and what is kept in the tree sequence)

    - Times of nodes and mutations, etc are optionally transformed from tskit units of backward time in generations into the time units used by the user to define models ("years before present", "years in the future from today", arbitrary time windows basically).
 
- If a SLiM tree sequence is spatial (either from slendr or pure SLiM), then locations are transformed either to the appropriate CRS used by the model in the [sf](https://r-spatial.github.io/sf/articles/sf1.html) format.

Taking care of the above is what most of the `!is.null(model)` conditions do -- non-slendr tree sequences don't have the additional metadata created by slendr itself (sampling tables with names of individuals to be recorded, coordinate reference systems, custom time units), but all the other SLiM tree sequence table munging is the same which is why I was able use the same code for this and only switch off the optional slendr-specific annotation code when needed.

Some of the points above (except the spatial stuff) applies also to slendr's msprime tree sequences (named individuals, custom times of nodes). There is currently exactly one function which has `msprime` in its name (`get_msprime_table_data()`) and that takes care of all of this. 

So even plugging in a "generic" tree sequence object should work without any changes to the internals of that function.

Is there a large-ish non-SLiM and non-msprime tree sequence file somewhere in tskit testing suite that I could borrow to test this on a non-simulated tree sequence?

---


**One thing I have to make very clear: the what I call 'raw' tree sequence tables (edges, nodes, individuals, mutations) are never touched by this R code. They are always accessible <u>in their original form</u>  via `ts_table(ts_object, "<edges|nodes|individuals|mutations>")`! None of the code munging you saw modifies those or adds additional information.** Most of the `get_slim_table_data` and `get_msprime_table_data` produces a more richly annotated tables useful for data analysis (with readable symbolic names of nodes of sampled individuals, model-specific times, geographic coordinates, etc). Those are available under `ts_nodes()` and `ts_edges()` in a rich annotated form (basically what can be directly used for visualization or statistics).

R/tree-sequences.R
bodkan: Yeah, it's not very pretty, is it? I was mostly sprinkling them here and there to see what slendr-specific code needs to be switched of or tweaked to work even without the slendr metadata (names of sampled individuals, coordinate reference systems, that sort of thing).

I will see if creating a different attribute type I'm attaching to a tree sequence R object could be used for this. Right now I have an attribute `source = <msprime|SLiM>` but based on the discussion above this is not actually a very meaningful bit of info for what's going on during the loading/processing.

bodkan: I have now refactored the code so that all those `is.null(model)` calls check to test if the currently processed/analyzed tree sequence are more readable. In a number of places I was able to join multiple of those conditions in a single place so the code is cleaner now too. The old version was quite ugly and it was all too clear I just injected them in multiple places as special case triggers.

R/tree-sequences.R
bodkan: Hmm, OK, I have to take that one statement back. My brain got confused there for a minute, mixing stuff up with the topic of [discussion](https://github.com/tskit-dev/tskit/discussions/2144) I originally started in the tskit forum on the +/- same topic. The workaround I implemented for getting the ape R phylo tree in the end doesn't have a problem with the "everyone is remembered" tree sequences anymore.

---

[... fast forward through Saturday + Sunday refactoring frenzy ...]

---

... also, as I was just about to send a justification for why I want/need that extra column "sampled", I re-read the documentation of tskit & pyslim yet again and discovered that the reason I wanted this extra column -- for the many situations in which a user wants to know which are **focal** individuals/nodes, not just remembered ones -- is actually captured by `TreeSequence.samples()`? Unless I'm missing something obvious, this is exactly what I basically re-implemented above.

I spent the whole weekend removing this, putting this back in, then renaming this column to `focal` because it fitted better to the semantics I had in mind for this... then I noticed that I'm probably reinventing the wheel (not realizing that remembered-retained-alive are not the only things that are tracked, but that a node can be remembered yet not be a sample).

**Just to make sure before I spend even more hours in the tskit rabbit hole: if a user simulates a tree sequence in which every single individual/node is remembered (let's say 10000 nodes are remembered out of 10000 nodes in the tree sequence total), then only simplifies down to a subset of 20 nodes, `ts.samples()` will only get them IDs of those 20 nodes, right? Every other node in that simplified tree sequence will be still flagged as `remembered` (probably less than 10000 we started with, but still likely quite many nodes), right?** Maybe @bhaller could help me here too? 

So, just to explain the code that probably did not make much sense to you @petrelharp (that `$sampled` column business), I think I was reinventing exactly what `ts.samples()` does. I wanted to give the user access to information about which nodes (out of possible many remembered) are the "focal nodes" they simplified to, and the chance to ignore the rest (even those they also explicitly remembered at first).

I'm astonished that even after 9 months of working with this I "discover" this "new feature". 🤦 I guess my excuse is that I have never used tree sequences outside of the slendr context...

bodkan: If the bold paragraph in the previous message is correct, then the `sampled` column would be simply populated by `TRUE` for whatever `node_id` is among `ts.samples()` and `FALSE` for any other node.

That is, unless I am -- for the 100th time -- missing something that's probably obvious to everyone else.

bhaller: > ...Maybe @bhaller could help me here too?

I'm afraid I cannot; you are already way beyond my level of expertise with tskit!  I mostly (but not entirely) understand how tree-sequence recording inside SLiM works.  My knowledge of other tskit things is fairly minimal.

> That is, unless I am -- for the 100th time -- missing something that's probably obvious to everyone else.

Don't beat yourself up, @bodkan – you're doing great and this stuff is amazing.  Learning to use a complex library like tskit always takes some iteration.

petrelharp: Right, ok, let's see:
1. everyone who is `remembered` will have their nodes marked as samples in the output tree sequence
2. everyone who is `retained` but not remembered will *not* be marked as samples
3. `simplify(x)` *only* leaves those nodes in `x` marked as samples but does not touch individual flags
4. there is no currently convenient way to "get individuals whose nodes are samples" although there **definitely should be**, we just haven't figured out the right API and done it yet

(Is this answering your question?  Maybe we should have some discussion on things like this on slack or even zoom - I don't always notice the github things right away.)

bodkan:
Hi folks, especially @bhaller and @petrelharp.

First of all, sorry for the delay with this PR. Things at `$DAY_JOB` went kind of sideways and I was forced to drop my work on slendr for more than a month. 

I’m finally back and thanks to your comments and suggestions I have now made the code much cleaner and more robust. I now have a reasonable version of the slendr R-tskit interface that can operate on standard non-slendr SLiM and msprime tree sequences.

This PR now accomplishes what I was aiming for and the new code is ready for testing by people who use SLiM/msprime (but not slendr) for simulation and who would like to check out slendr’s features for tree-sequence analysis. I’ve already had a number of people express interest in this, so the sooner I can get things into their hands the better. Once the GitHub Actions CI unit tests pass (currently in progress), I will merge the PR.

---

In case you remember what the various issues were, then very briefly (if you don’t, just ignore this, I'm logging this also for my own benefit):

- I am no longer reinventing the wheel with the `sampled/focal` column I’ve been adding to the slendr node/individual table. I wasn’t aware of the existence of `TreeSequence.samples()` which does the same thing so I'm using that where applicable.
- `ts_load()` and friends no longer treat tree sequences as coming from either SLiM or msprime and ignoring everything else with a warning (which was the case in the original version of this PR). Now, tree sequence is either coming from SLiM, or its regarded as "generic" tskit tree sequence (msprime tree sequence or other source). This refactoring made the internal code much cleaner with less awkward special case conditions.
- There is a new vignette which expands on the overview from the [first comment of this PR](https://github.com/bodkan/slendr/pull/91#issuecomment-1114329598). As discussed, perhaps @bhaller and I could talk about which (if any) of this stuff would be useful to show in the SLiM manual. I kept this vignette extremely short, because everything in it is described in all other vignettes in much greater detail. The point of the new vignette is to show "you can use SLiM and msprime, then use standard slendr functions to load/process/analyze tree sequences". It will be available [here](https://www.slendr.net/articles/vignette-08-nonslendr-tskit.html) once the PR is merged.
- I added a couple of rudimentary unit tests (such as [this one](https://github.com/bodkan/slendr/blob/7c45135b4afe4eceea8e9eee2f817a7ad9be75f4/tests/testthat/test-pure-slim-vs-slendr.R) and [this one](https://github.com/bodkan/slendr/blob/28218ea1b6ef2ed9b03ea32c3871bceb9f48b18a/tests/testthat/test-pure-msprime-vs-slendr.R)) which run a simple pure SLiM or msprime script and also a slendr version of the same (very simple) model and compare the resulting tree-sequence tables and trees. Additional unit tests will be added if/when issues arise by people who want to use slendr for data analysis (but not necessarily simulation).

---

I have talked to a bunch of people recently who use SLiM for awesome spatial selection work (range expansion surfing, etc.) — they are not that interested in slendr as a simulation tool (they need much more detailed control over spatial dynamics than what slendr provides) but they are very interested in the spatial R-tskit interface of slendr. I have also had two students working on slendr-adjacent things who rely on this PR to be merged.

Because of that, as I mentioned above, I will merge this PR as soon as I know that the GitHub CI checks and unit tests pass. If there are corner cases that I have not taken into account, the brave users who reached out to me offered to help me fix those (the only thing needed will be a SLiM script that produces a tree sequence that fails to load/process/analyze). That said, given that slendr tree sequences *are* normal tskit tree sequences, I don’t expect dramatic problems. The only thing the current implementation of R-tskit layer is doing is that it’s ignoring slendr-specific metadata when that metadata is missing, and uses the same code for non-slendr tree sequences otherwise.

There are other updates too, but I will leave those to an email update that I’ll send soon.

Thanks again for your input!

bodkan:
My goodness. 4 hour GitHub Actions CRAN checks. (Not running the tests on Windows because I don't have an access to a Windows machine with an R development environment set up. But given that SLiM now runs on Windows without issues, it's about time I worked on this at some point soon.)

Merging the PR now and I will also tag a new version.

Currently working on the first CRAN submission. Wish me luck, if past experiences are any indication, I'm going to need it...

bodkan:
Hello @bhaller 

> Happy to talk about adding a slendr-based recipe to the SLiM manual. It could go in as section 17.11, if having at added to the end makes sense to you, or somewhere internal to chapter 17.

I will take a closer look and let you know. 👍 

> The description of slendr in section 1.10 could also probably use revision, given these changes.

Absolutely, I will take a look and suggest adjustments where necessary.

> I'm a bit swamped at the moment – I'm at a workshop thing for the next 3.5 weeks that pretty much takes up all my time. If you'd like to propose changes/additions, I'll be happy to add them; if you want to have a zoom to discuss first, that will wait at least 1.5 weeks, very possibly longer, for a more quiet time in my life. :->

Same, the only reason I had time for all this is that I dropped everything else unconditionally to push the first release to CRAN and submit the paper, now that lots of people tried slendr and no major disaster has happened. I don't think I'll have a lot of time for those changes/additions over the next 2-3 weeks either. But I will be in touch sometime in July. (Even if I manage earlier, there's no pressure.)

> One question: is slendr now SLiM 4 compatible? I hope so. The changes needed ought to be fairly minimal, and with a little elbow grease it should be possible to have your scripts run on both 3.7 and 4.0 (let me know if you need any advice on that). It would be great if the SLiM 4 release, which I plan to do in early to mid August, did not break slendr and create a crisis. :->

I had absolutely no time to test things with SLiM 4 just yet :( But it is important thing to check, so I opened an issue (#98) and will make sure to test this soon. Would you say that running the current slendr unit tests against the SLiM 4 binary be a good start? I.e. that only a few syntactic/semantic would be needed that I could fix a few errors that might arise? In any case, I will post potential issues there and ping you when necessary. Thanks for bringing this up.

> Exciting! I'm curious who you've been talking to about range expansion surfing, etc. :->

In this particular instance mostly Excoffier people, so not that surprising. :)

bhaller:
> ...Same, the only reason I had time for all this is that I dropped everything else unconditionally to push the first release to CRAN and submit the paper, now that lots of people tried slendr and no major disaster has happened. I don't think I'll have a lot of time for those changes/additions over the next 2-3 weeks either. But I will be in touch sometime in July. (Even if I manage earlier, there's no pressure.)

Sounds good.

> > One question: is slendr now SLiM 4 compatible? I hope so. The changes needed ought to be fairly minimal, and with a little elbow grease it should be possible to have your scripts run on both 3.7 and 4.0 (let me know if you need any advice on that). It would be great if the SLiM 4 release, which I plan to do in early to mid August, did not break slendr and create a crisis. :->
> 
> I had absolutely no time to test things with SLiM 4 just yet :( But it is important thing to check, so I opened an issue (#98) and will make sure to test this soon. Would you say that running the current slendr unit tests against the SLiM 4 binary be a good start? I.e. that only a few syntactic/semantic would be needed that I could fix a few errors that might arise? In any case, I will post potential issues there and ping you when necessary. Thanks for bringing this up.

I expect it to only be syntactic tweaks, yes.  By and large, SLiMgui will offer to "autofix" the issues for you if you open a generated slendr script in SLiMgui and run it; each error encountered should produce a new "autofix" suggestion.  Accepting those fixes will get you a working SLiM 4 script (although there might be a case or two where autofix is not sufficient).  The SLiM 4 beta has links to more information about exactly what changed.  The thing that might bite you the most is that sim.generation no longer exists; it is now community.tick or sim.cycle, to get the same sort of information (see the docs regarding what the difference between those two things is).  If you use it a lot, you might find it simpler to just make a user-defined function named `time()` that checks the version number and returns the correct property, or something.  Sorry about that; I had my reasons, but the changeover has created a lot fo work for me, too.  :->

> > Exciting! I'm curious who you've been talking to about range expansion surfing, etc. :->
> 
> In this particular instance mostly Excoffier people, so not that surprising. :)

:->

PR #102, opened 2022-08-07
Windows port of the msprime functionality

bodkan:
A student in the department (complete newbie to popgen and simulations) has a Windows laptop and wants to run some simple coalescent simulations. This should hopefully unlock the option to run Python-based msprime simulations in slendr.

Running SLiM in slendr requires working with the OS command-line and slendr currently does it using unix terminal under the hood. Porting this to Windows will be more work (it's been > 10 years since I touched `cmd.exe` on a Windows machine), but at least the Windows Python/msprime port should be relatively straightforward, so let's focus on that here.

### DIFF
--- a/R/compilation.R
+++ b/R/compilation.R
@@ -635,7 +635,7 @@ msprime <- function(model, sequence_length, recombination_rate, samples = NULL,
     cat("--------------------------------------------------\n\n")
   }
 
-  reticulate::py_run_string(sprintf("import os; os.system('%s')", msprime_command))
+  reticulate::py_run_string(sprintf("import os; os.system(r'%s')", msprime_command))
 
   # if (system(msprime_command, ignore.stdout = !verbose) != 0)
   #   stop("msprime simulation resulted in an error -- see the output above", call. = FALSE)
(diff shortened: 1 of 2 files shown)


bhaller:
You might consider just recommending that people do this sort of thing under the WSL when on Windows.  SLiM runs under the WSL just fine, including SLiMgui; chapter 2 of the manual has install instructions for that configuration.  But of course if it worked on native Windows too, that would be nice.

bodkan:
Yeah, that is definitely a good option. I've been hesitant to recommend something like that officially, simply because I have no way of testing things myself. In fact, I don't even have access to a Windows machine. Perhaps I should finally look for a reasonable/free virtualization option for macOS.

I will probably wait for someone to _really_ need this and then work with them to sort a full Windows port together. It would be nice to have the possibility to use slendr/SLiM without WSL because that would make the package truly multiplatform.

bhaller:
> Yeah, that is definitely a good option. I've been hesitant to recommend something like that officially, simply because I have no way of testing things myself. In fact, I don't even have access to a Windows machine. Perhaps I should finally look for a reasonable/free virtualization option for macOS.
> 
> I will probably wait for someone to _really_ need this and then work with them to sort a full Windows port together. It would be nice to have the possibility to use slendr/SLiM without WSL because that would make the package truly multiplatform.

This is much the same position I am in with SLiM; I have no Windows machine, and no desire to learn how to do software development on Windows, so I waited until someone came along who wanted to volunteer to do the port.  :->

Issue #97, opened 2022-06-21
Add option to skip serialization of slendr configuration files


bodkan:
With the awesome work that @juledrewa has been doing on automated model generation that will be merged into the main branch soon, I think it makes sense to add the possibility to skip the serialization of compiled slendr files to disk (which can be costly in terms of time and space, especially for many many models compiled in parallel), and simply create an R slendr model object that's not backed by config files.

One idea we had was to try to implement a method for quick model fitting via msprime simulation. Unlike the SLiM back end, with Python we have the advantage to call the msprime slendr backend directly from R, without having to go through the command-line and config files. For these use cases, compiling files to disk is not necessary. Doing simple `model <- compile_model(..., serialize = FALSE)` or such would be nice because we could go directly to `msprime(model, ...)` without having to compile a bunch of files to disk first (which could be annoying for thousands/millions of models that might have to be compiled).

Just making a note of that so that I don't forget, and also so that @juledrewa and @FerRacimo know about this.

bodkan:
Actually, in this situation `msprime(model, ...)` could directly return an R-based tskit tree sequence object, without having to save it to disk first in order to be loaded by `ts_load()` after the simulation...

bodkan:
Closed [here](https://github.com/bodkan/slendr/pull/112).

PR #112, opened 2022-09-05
Make serialization of slendr models to disk optional (for faster msprime coalescent runs)

bodkan:
This is is an attempt to implement #97. Skipping serialization to disk makes it possible to call our msprime back-end Python code directly, without having to go through the standard _Bunch of Files on Disk_ (TM) format of slendr models (normally needed to call the SLiM back-end script on the command line). This will, in turn, make coalescent simulations even faster, opening up the possibility to develop model fitting procedures in other projects (that functionality won't be part of slendr).


bodkan:
Things are looking good so far. The only thing necessary was to make the back-end msprime simulation code into its own function. Then, `compile_model` was given a `serialize = TRUE | FALSE` argument (default `TRUE`), which allows skipping the writing of model configuration files to disk. The `msprime()` function can detect this and simply use R data frames (converted to pandas DataFrames) containing the model configuration data stored in memory (not on disk) and plug them into the msprime Python coalescent script.

More unit tests are needed to make sure I didn't miss some combination of missing gene-flow events, now resize events, etc. -- this would make some of those data.frames NULL/None, which needs to be take care of accordingly.

PR #120, opened 2023-01-18
Fix undesirable artefacts during map zooming/cropping for some CRS

### LINKED ISSUE #116: Strange things are happening with custom CRS projections in some corners of the world
I've been playing with spatial models in other corners of the world than just Eurasia, also exploring different area-specific coordinate reference systems. Discovered weird bugs with plotting code which probably indicate something is off with the internal sf-based code which handles transformations of spatial geometric objects.

Consider zooming in on South America here using slendr's `world` function:

```
> map <- world(xrange = c(-90, -20), yrange = c(-58, 15))

OGR data source with driver: ESRI Shapefile 
Source: "/private/var/folders/d_/hblb15pd3b94rg0v35920wd80000gn/T/RtmpLmReLu/naturalearth", layer: "ne_110m_land"
with 127 features
It has 3 fields

Warning message:
No explicit coordinate reference system (CRS) was specified.
A default geographic CRS (EPSG:4326) will be used but please make
sure this is appropriate for your geographic region of interest.
```

We don't specify a coordinate reference system, and get a warning as expected, but otherwise there's no problem, and we can plot the map with slendr as such:

```
plot_map(map)
```

image

However, when a customized projection system is specified, the plotted map is weird:

```
map <- world(xrange = c(-90, -20), yrange = c(-58, 15), crs = "EPSG:31970")
plot_map(map)
```

image

What's the grey bar on the right? Is that part of the map object? Plotting artifact?

Now, I picked the CRS `EPSG:31970` simply googling "best CRS for South America", so I'm not claiming this is the best CRS projection available... but the map is clearly off. 

Serveral things could be wrong here:

1. slendr's internal sf-based cropping/zooming/transformation
2. building and zooming of the map works, but it's the slendr's plotting that's broken
3. ?

bodkan:
This PR fixes [issue #116](https://github.com/bodkan/slendr/issues/116). In particular, it reverses the order of `CRS transformation -> cropping`, into `cropping -> CRS transformation` as explained [here](https://github.com/bodkan/slendr/issues/116#issuecomment-1383848371).

This results in minor changes in the appearance of transformed maps (i.e. maps which use a projected CRS instead of the WGS-84/EPSG:4326 geodesic lat/lon coordinate system). The changes are, however, purely aesthetic and not functional. For an example see [this](https://github.com/bodkan/slendr/issues/116#issuecomment-1327879958).


bodkan:
Plots in the vignette with examples from the slendr paper needed minor cosmetic updates. All done now.

PR #121, opened 2023-01-31
Allow for the possibility to have non-spatial populations in an otherwise spatial model

bodkan:
It might be (?) occasionally (?) useful (?) to have populations in a spatial model which do not have a well-defined spatial boundary. For instance, if we don't know or don't care about the spatial location of some population in the model (but we do know this or care about this for other populations), or if we want to have a "normal" coalescent non-spatial model, but want to some populations in that model annotated with some spatial metadata (not implemented yet, but will be soon). Such models were not valid in the original implementation of slendr.

It turned out that the changes to the compilation process of slendr models and the SLiM back end required to make this possible turned out to be quite minor. This PR implements those changes and adds a new example to the "semi-spatial" demes-based simulations vignette demonstrating this possibility.


bodkan:
Unfortunately, yet another conda breakage occurred while working on this. Took forever to fix things, but at least I used the opportunity to bump up required versions of msprime/tskit/pyslim and Python (3.11) to their latest releases. Again, all of this is a purely internal matter as far the slendr is concerned. Our `setup_env()` function takes care of everything automatically, even for users who don't even have Python on their systems ([reticulate](https://rstudio.github.io/reticulate/) really is incredible).

slendr is now ready for its v0.5 release to CRAN, yay!

PR #122, opened 2023-02-02
Use NOT_CRAN=false on Windows-based GitHub Actions

bodkan:


bodkan:
For some reason, Windows-based CRAN check on R-devel started throwing the following warning:

```
GDAL Message 1: The definition of geographic CRS EPSG:4258 got from GeoTIFF keys is not the
same as the one from the EPSG registry, which may cause issues during reprojection operations.
Set GTIFF_SRS_SOURCE configuration option to EPSG to use official parameters (overriding the
ones from GeoTIFF keys), or to GEOKEYS to use custom values from GeoTIFF keys and drop the
EPSG code.
```

There is zero usage of EPSG:4258 in the codebase. There's no GeoTIFF of any kind being used in slendr. Given that this only happens on a particular Windows setup used by CRAN, this PR attempts to re-activate a Windows-based GitHub Actions CI pipeline that is close to CRAN setup. (I didn't do Windows checks here because there used to be no slendr on Windows.)

For some reason that I still don't understand, default devtools checks (and, by extension, GitHub Actions R CI checks) set `NOT_CRAN=true`. This PR reverts that so that I can use GitHub Actions to do Windows checks more in line with what CRAN actually does.

PR #123, opened 2023-03-02
New ts_ibd function

bodkan:
This adds a new function -- _slendr_'s interface to [`TreeSequence.ibd_segments()`](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.TreeSequence.ibd_segments).

As explained in the `?ts_ibd` manpage, this is not a real wrapper. R handles heavy iteration extremely poorly so the [documented](https://tskit.dev/tskit/docs/stable/ibd.html) use cases wouldn't really work here. Certainly not for large tree sequences.

Instead, `ts_ibd()` collects all requested IBD data (either all individual IBD segments when `coordinates = TRUE` or counts and total pairwise IBD amount when `coordinates = FALSE`, which is the default) and returns the results as a plain data frame (EDIT: for spatial tree sequences the returned IBD table is now fully spatially annotated and is of the _sf_ data type).

To help to make things manageable, pruning the IBDs to be returned either by setting the minimum length of an IBD segment to be considered, or via setting the maximum age of an ancestor of an IBD pair, is still supported. In fact, given how easy it is to choke on too much IBD, `ts_ibd()` writes a warning message if _all possible IBDs_ are being requested by the user (something that is most likely an oversight during normal data analysis).

Similarly, the `within = ` and `between = ` arguments are also supported. In line with the rest of the _slendr_ `ts_*()` library, these arguments accept symbolic names of individuals, not just integer IDs of nodes.


bodkan:
IBD tracts collected from spatial tree sequences are now annotated with spatial coordinates of nodes and returned as spatial _sf_ objects by default.

---

As an aside, note that although `ts_ibd()` returns IBD data in a tabular format as mentioned in the first post, and doesn't work with iteration (and never will), if users _need_ to do iterate over massive amounts of IBD, they can always use the _reticulate_-d iteration just like is shown in [tskit docs](https://tskit.dev/tskit/docs/stable/ibd.html#data-structures). Honestly though, at that point it's probably better to use Python.

Issue #129, opened 2023-04-03
Improve robustness of the "simulation finished" check in `slim()`


bodkan:
@IsabelMarleen has encountered an interesting curiosity while implementing a slendr importer from [Demes](https://popsim-consortium.github.io/demes-docs/latest/introduction.html). When she imports the Gutenkunst _et al._ example from the [Demes repository](https://github.com/popsim-consortium/demes-spec/blob/main/examples/gutenkunst_ooa.yaml), successfully generating a slendr model object, she runs into an interesting issue when she simulates it with our built-in `slim()` function.

Apparently it turned out to be a challenge to produce a minimum reproducible example, but she provided a serialized, compiled slendr model that reproduces this: [gutenkunst_slendr_model.zip](https://github.com/bodkan/slendr/files/11137640/gutenkunst_slendr_model.zip). Briefly, loading her serialized model first:

```
library(slendr) # v0.5.1
init_env()

model <- read_model("<path to a directory with the unzipped model data>")
plot_model(model)
```

It seems that Isabel's demes-to-slendr importing code works as it should (let's ignore the poor plotting of slendr for now):
<img width="893" alt="image" src="https://user-images.githubusercontent.com/16516593/229494037-5cee3dad-4002-4bf8-b6cc-6cd939a630d2.png">

However, running the model with `slim()` gives us:

```
> ts <- slim(model, sequence_length = 1000, recombination_rate = 0, verbose = TRUE)
Tree-sequence recording is on but no sampling schedule was given. Generating one for all individuals surviving to the end of the simulation.
--------------------------------------------------                            
SLiM command to be executed:

slim   \
    -d 'SAMPLES="/var/folders/d_/hblb15pd3b94rg0v35920wd80000gn/T//RtmpY74O5c/file79e945962461"' \
    -d 'MODEL="/Users/mp/Downloads/gutenkunst_slendr_model"' \
    -d 'OUTPUT_TS="/var/folders/d_/hblb15pd3b94rg0v35920wd80000gn/T//RtmpY74O5c/file79e962919bf5.trees"' \
    -d SPATIAL=F \
    -d SEQUENCE_LENGTH=1000 \
    -d RECOMB_RATE=0 \
    -d BURNIN_LENGTH=0 \
    -d SIMULATION_LENGTH=8800 \
    -d 'OUTPUT_LOCATIONS=""' \
    -d COALESCENT_ONLY=T \
    -d MAX_ATTEMPTS=1 \
[...]

PR #130, opened 2023-04-07
Fixes for minor issues discovered during work on demografr and other things

bodkan:
A few (minor) issues have been discovered during the work on the ABC R package [_demografr_](https://github.com/bodkan/demografr/) as well as through in-person reports from users. Turns out throwing thousands of randomized models and models written by other people not involved with slendr at the package means it gets tested much more extensively! 🔥 :)


bodkan:
@IsabelMarleen as discussed on Slack, I'm merging this now. Please re-run `devtools::install_github("bodkan/slendr")` to get the development version from the `main` branch to keep things in sync with your thesis code.

Also, it would be great if you could test what happens with your Gutenkunst slendr model generated by your `import_demes()` function as reported [here](https://github.com/bodkan/slendr/issues/129). I suspect changes implemented in this PR won't make a difference there but still... worth a try? Regardless of the outcome, could you please post an update note under that issue? Thank you. (**Not urgent!**)

Issue #132, opened 2023-04-20
Incorrect error messages in `gene_flow()` caused by time errors


IsabelMarleen:
There is an issue with the error messages in the gene_flow() function for time inconsistencies. I found a minimal example where an error caused by an inconsistency with the assumed direction of time causes an error about the populations not being present in the time window and a second scenario where the opposite happens. This seems like an issue that could be caused by a condition being the wrong way around.

```
# The model
anc <- population("ancestor", time = 9e6, remove=8e6, N=100)
popA <- population("A", time = 8e6, parent=anc, N=100)
popB <- population("B", time = 7e6, parent=anc, N=100)
```

`gf <- gene_flow(from = popA, to = popB, start = 4e6, end = (4e6+5), rate = 0.3)`
This throws the error:

> Both A and B must be present within the gene-flow window 4e+06-4000005 

but it should complain that the end time is before the start time.

`gf <- gene_flow(from = popA, to = popB, start = 8e6, end = (8e6-5), rate = 0.3)`
This throws the error:

> Specified times are not consistent with the assumed direction of time (gene flow A -> B in the time window 8e+06-7999995)

but it should complain that population B does not exist yet in this time window.

bodkan:
Thanks for the report.

Hmm, I am sort of aware of this issue. I have poked around this a bit and discovered some weird edge cases in which the `gene_flow()` function does not have a complete information about the intended flow of time... which doesn't appear to be the case here in your minimal example (thanks for providing this, this is really great).

I think it had something to do with one population existing _within_ the gene-flow time window, but the other existing outside (either before it or after it)? I'll have to check my notes.

The easiest thing we could do for now is simply write a generic error about a gene flow being somehow broken. Less satisfying and less informative, but at least not misleading.

bodkan:
Some notes will follow as I'm working through this.

Consider this model:

```
popA <- population("A", time = 1, N = 100)
popB <- population("B", time = 5, N = 100)

gf <- gene_flow(from = popA, to = popB, start = 10, end = 20, rate = 0.3)
```

The gene flow encoded by that code clearly implies forward-time direction (start at 10 -> end at 20), but the populations themselves do not encode any information. They are two ancestral populations, but without the information from the gene flow (which implies forward time), we can't tell which sort of time direction should they involve.

The direction is only 100% clear in `compile_model()`. For some models (like the one in this example), the populations themselves do not add additional information.

[...]

PR #134, opened 2023-04-21
Fixing various gene flow errors

bodkan:
This finally addresses inconsistent error messages formally reported in #132.

Also simplifies and cleans up the `gene_flow()` code a bit in the process. No breaking changes or changes to the user-facing interface introduced by this (but I hope some yet-undiscovered issues will be correctly caught as a result of this).


bodkan:
@IsabelMarleen: Still have to do more testing but I'm afraid it's already clear we will have to do the _"gene flow happens earliest at +1 (or -1, for backwards time models) time point post split"_ as we discussed a couple of times.

For your demes project, issue a warning wherever an adjustment of this kind needs to be made during compilation from demes into slendr.

(This is *not* a priority though! Unless that's something that absolutely has to happen urgently, let's deal with this fully after your thesis is submitted. Focus on your 🐘 now. :))

Issue #136, opened 2023-05-04
There's an issue with saving and/or loading a simplified tree sequence (carrying slendr metadata)


bodkan:
Let's encode the following trivial model:

```
model <-
  population("pop", time = 1, N = 100) %>%
  compile_model(generation_time = 1, simulation_length = 1000, direction = "forward")
```

Simulate a tree sequence from it, save that tree sequence to disk:

```
ts <- slim(model, sequence_length = 1000, recombination_rate = 0)
ts_save(ts, "/tmp/big.trees")
```

So far so good.

- Loading the tree sequence (plus model data) works:

```
> ts_load("/tmp/big.trees", model)

╔════════════════════════╗
║TreeSequence            ║
╠═══════════════╤════════╣
║Trees          │       1║
╟───────────────┼────────╢
║Sequence Length│    1000║
╟───────────────┼────────╢
║Time Units     │   ticks║
╟───────────────┼────────╢
║Sample Nodes   │     200║
╟───────────────┼────────╢
║Total Size     │95.3 KiB║
╚═══════════════╧════════╝
╔═══════════╤════╤════════╤════════════╗
║Table      │Rows│Size    │Has Metadata║
╠═══════════╪════╪════════╪════════════╣
║Edges      │ 366│11.4 KiB│          No║
╟───────────┼────┼────────┼────────────╢
║Individuals│ 256│26.8 KiB│         Yes║
╟───────────┼────┼────────┼────────────╢
║Migrations │   0│ 8 Bytes│          No║
╟───────────┼────┼────────┼────────────╢
║Mutations  │   0│ 1.2 KiB│          No║
╟───────────┼────┼────────┼────────────╢
║Nodes      │ 367│14.3 KiB│         Yes║
╟───────────┼────┼────────┼────────────╢
║Populations│   1│ 2.3 KiB│         Yes║
╟───────────┼────┼────────┼────────────╢
║Provenances│   1│33.9 KiB│          No║
╟───────────┼────┼────────┼────────────╢
║Sites      │   0│16 Bytes│          No║
╚═══════════╧════╧════════╧════════════╝
```

- Simplifying the tree sequence and _then_ loading it also works:

```
> ts %>% ts_simplify() %>% ts_save("/tmp/small1.trees")
> ts_load("/tmp/small1.trees", model)
╔════════════════════════╗
║TreeSequence            ║
╠═══════════════╤════════╣
║Trees          │       1║
╟───────────────┼────────╢
[...]

PR #137, opened 2023-05-08
Fix saving / loading of tree sequences after simplification to a subset of sampled individuals

bodkan:
An attempt at fixing #136.


bodkan:
Merging now. From the changelog:

> Fix for `ts_load()` failing to load _slendr_-produced tree sequences after they were simplified down to a smaller set of sampled individuals (reported [here](https://github.com/bodkan/slendr/issues/136)). The issue was caused by incompatible sizes of the sampling table (always in the same form as used during simulation) and the table of individuals stored in the tree sequence after simplification (potentially containing a smaller set of individuals than in the original sampling table). To fix this, _slendr_ tree sequence objects now track information about which individuals are regarded as "samples" (i.e. those with symbolic names) which is maintained through simplification, serialization and loading, and used by _slendr_'s internal machinery during join operations. (PR [#137](https://github.com/bodkan/slendr/pull/137))

PR #144, opened 2023-10-14
Support for manually created tree sequences with spatial information

bodkan:
While doing some follow up inference work, it turned out that slendr's `ts_load()` function makes an implicit assumption that the only spatial tree sequences (those which store spatial coordinates in the individual tree sequence table) are those coming from SLiM. Naturally, loading a manually created tree sequence (a tree sequence composed from individual tables) ignored spatial coordinate columns.

So, with a tree sequence created like this:

```
nodes <- tribble(
  ~id, ~is_sample, ~time, ~individual,
  0,   1,          0,     0,
  1,   1,          0,     1,
  2,   0,          3,     2,
  3,   1,          3,     3,
  4,   0,          6,     4
)

edges <- tribble(
  ~left, ~right, ~parent, ~child,
  0,     10,     2,       0,
  0,     10,     2,       1,
  0,     10,     4,       3,
  0,     10,     4,       2

)

individuals <- tribble(
  ~id, ~flags, ~x,  ~y,
  0,   0,      2,   4,
  1,   0,      8,   7,
  2,   0,      5,   5.5,
  3,   0,      4,   2,
  4,   0,      4.5, 3.75
)

ts <- compose_ts(nodes, edges, individuals)
```

calling `ts_nodes(ts)` on a serialized and `ts_load()`-ed tree sequence returns this (note the absence of coordinates of nodes):

image

**This PR is a mildly terrifying attempt at a few surgical cuts deep into the slendr internals to make `ts_load()` load spatial columns even for such tree sequences without breaking anything else.**

With changes introduced by this PR, `ts_load()` and subsequent `ts_nodes(ts)` do what they're supposed to (note the presence of the location column as well as the fact that the data is now in the spatial _sf_ data format):

image

The question now is, did something else break in the process? Guess we will find out!


bodkan:
Oh, while I'm at this -- I had to turn off the testthat pipeline because of sudden timeout issues on GitHub Actions' part. I should try to switch this back up, to see if things have been fixed in the meantime.

Issue #171, opened 2025-04-15
Proposed fix for the upcoming release of rnaturalearth


PMassicotte:
Hi there.

We are planing a new release of the `rnaturalearth` package. naturalearth vecotr files will now be converted/downloaded as `gpkg` format:

https://github.com/ropensci/rnaturalearth/blob/c0a1fa2e99f6a6597bc6dbf763b0a970f56666c7/NEWS.md?plain=1#L3-L19

I have noticed that this will break some parts of this package where you manually downloaded the `.zip` file directly from naturalearth and then use `ne_load()`.

https://github.com/bodkan/slendr/blob/0a9ad4973a826a207404b1975eb94cc46fb29af3/R/interface.R#L798-L802

https://github.com/bodkan/slendr/blob/0a9ad4973a826a207404b1975eb94cc46fb29af3/R/interface.R#L810-L814

Would it be possible to adjust the code above to use `ne_download()` first and then use `ne_load()`?

For example,

```r
utils::download.file(
  url = sprintf(
    "https://naturalearth.s3.amazonaws.com/%sm_physical/%s",
    size,
    file
  ),
  destfile = path,
  quiet = TRUE
)
utils::unzip(path, exdir = ne_dir)
```

Could be replaced with

```r
rnaturalearth::ne_download(
  scale = size,
  type = "land",
  category = "physical",
  destdir = ne_dir,
  load = FALSE
)
```

And then loaded with

```r
rnaturalearth::ne_load(
  scale = size,
  type = "land",
  category = "physical",
  destdir = ne_dir
)
```

There is also this part that would need to be updated:

https://github.com/bodkan/slendr/blob/0a9ad4973a826a207404b1975eb94cc46fb29af3/logo.R#L11-L19

You can try to use the dev version of `rnaturalearth` to make sure it works fine.

With regards,
Phil

bodkan:
Hi Phil,

Thanks for much for the heads up about the upcoming changes, and even more for actually crawling through my code and proposing a solution. I love the R community.

I've been traveling quite a bit for some teaching and meetings but I will take a look at this over the next couple of days.

Cheers,
Martin

P. S. Also, a few years late, but thanks for stepping up to take care of `rnaturalearth` back in the day. I really appreciate everything you folks have been doing to keep the lights on.

PMassicotte:
Thank you very much for you kind message. There is no rush at all. :)

PR #173, opened 2025-05-05
Fix for the upcoming release of rnaturalearth

bodkan:
See here: https://github.com/bodkan/slendr/issues/171

### DIFF
--- a/R/interface.R
+++ b/R/interface.R
@@ -788,14 +788,12 @@ world <- function(xrange, yrange, landscape = "naturalearth", crs = NULL,
     # the small scale Natural Earth data is bundled with slendr
     ne_dir <- file.path(tempdir(), "naturalearth")
     if (scale == "small") {
-      utils::unzip(system.file("naturalearth/ne_110m_land.zip", package = "slendr"),
-                   exdir = ne_dir)
+      utils::unzip(system.file("naturalearth/ne_110m_land.zip", package = "slendr"), exdir = ne_dir)
     } else {
-      size <- ifelse(scale == "large", 10, 50)
-      rnaturalearth::ne_download(scale = size, type = "land", category = "physical", destdir = ne_dir, load = FALSE)
+      rnaturalearth::ne_download(scale = scale, type = "land", category = "physical", destdir = ne_dir, load = FALSE)
     }
 
-    map_raw <- rnaturalearth::ne_load(scale = size, type = "land", category = "physical", destdir = ne_dir)
+    map_raw <- rnaturalearth::ne_load(scale = scale, type = "land", category = "physical", destdir = ne_dir)
     sf::st_agr(map_raw) <- "constant"
 
     # define boundary coordinates in the target CRS
--- a/R/interface.R
+++ b/R/interface.R
@@ -792,26 +792,10 @@ world <- function(xrange, yrange, landscape = "naturalearth", crs = NULL,
                    exdir = ne_dir)
     } else {
       size <- ifelse(scale == "large", 10, 50)
-      file <- sprintf("ne_%sm_land.zip", size)
-      if (!dir.exists(ne_dir)) dir.create(ne_dir)
-      path <- file.path(ne_dir, file)
-      utils::download.file(
-        url = sprintf("https://naturalearth.s3.amazonaws.com/%sm_physical/%s", size, file),
-        destfile = path, quiet = TRUE
-      )
-      utils::unzip(path, exdir = ne_dir)
+      rnaturalearth::ne_download(scale = size, type = "land", category = "physical", destdir = ne_dir, load = FALSE)
     }
 
-    # TODO: this function uses internally rgdal, which is to be retired by 2023
-    # silence the deprecation warning for now, as it would only confuse the user
-    # and figure out a way to deal with this (either by providing a PR to the devs
-    # or hacking our own alternative)
-    suppressWarnings(
-      map_raw <- rnaturalearth::ne_load(
-        scale = scale, type = "land", category = "physical",
-        returnclass = "sf", destdir = ne_dir
-      )
-    )
+    map_raw <- rnaturalearth::ne_load(scale = size, type = "land", category = "physical", destdir = ne_dir)
     sf::st_agr(map_raw) <- "constant"
 
     # define boundary coordinates in the target CRS


bodkan:
I think the failed SLiM runs on Windows are related to this error which popped up:

```
C:\Users\RUNNER~1\AppData\Local\R-MINI~1\envs\PYTHON~1.2\Lib\site-packages\pyslim\slim_metadata.py:874: UserWarning: This tree sequence is not the current SLiM format, so some operations may not work. Use `pyslim.update( )` to update the tree sequence.
```

No idea where this comes from though. Debugging Windows-related errors on GitHub Actions servers is next to impossible.

I will have to fire up Windows VM locally and investigate.

bhaller:
> I think the failed SLiM runs on Windows are related to this error which popped up:
> 
> ```
> C:\Users\RUNNER~1\AppData\Local\R-MINI~1\envs\PYTHON~1.2\Lib\site-packages\pyslim\slim_metadata.py:874: 
> UserWarning: This tree sequence is not the current SLiM format, so some operations may not work.
> Use `pyslim.update( )` to update the tree sequence.
> ```
> 
> No idea where this comes from though. Debugging Windows-related errors on GitHub Actions servers is next to impossible.
> 
> I will have to fire up Windows VM locally and investigate.

Just a warning, not an error, right?

I believe this is pyslim complaining that it doesn't understand the SLiM 5 metadata format.  For your purposes, I believe it should work fine if you just ignore the warning.  @petrelharp is working on a new pyslim version for SLiM 5.

bodkan:
Oh crap, you’re right. Embarrassing!

In my defense, yesterday was a horrible day at work and I wrote these comments as reminders for myself when I pick this up again, because I was forced to drop this halfway through… :( I was not paying much attention as I had too much adrenaline flowing around my body. :D

There is _something_ going on Windows which I will figure out. I just have to remember how to set up this msys2 thing and such. 😅 Also, even a tiny little thing on VM’d Windows takes forever to run, making debugging of anything take at least a day. :)

Thanks again, @bhaller! As always, I appreciate you keeping an eye on what I’m trying to do here.

bhaller:
Rule 1 of life: avoid Windows if possible.  I hope your work situation improves!  :-O

bodkan:
OK, a couple more detailed notes to help me remember where to pick this up when I come back after vacation.

1. Rather than work on a barely-functional Windows setup in a VM, I went all in, upgraded to SLiM 5 on my main machine (macOS). Figured that if things break there (yes they do), the problem is not Windows-specific and would need fixing anyway... and, after fixing it, the issue would be resolved also on Windows.

2. Interestingly, after upgrading to SLiM 5 and all the recent Python dependencies (msprime, tskit, tspop, etc.), **all unit tests in slendr (about 1100 tests, including SLiM-specific unit tests) run successfully -- except one, one of the tests which runs SLiM** (I suspect this is the one that originally triggered failing tests on GitHub Actions Windows servers). This suggests that there's something very interesting going on. I didn't try rendering vignettes yet, because that would make the whole testing procedure run several times longer.

3. Unfortunately, creating a minimum reproducible example is tricky, because the SLiM engine script built into slendr is almost 1000 lines long and is very complex. In the past I did chisel it down to only a tiny minimum SLiM script, but that was extremely difficult and time-consuming. For the time being, I'll try to dig into the script as it is, because it looks like the problem is very very specific.

4. The failing unit test script is [this one](https://github.com/bodkan/slendr/blob/rnaturalearth-fix/tests/testthat/test-time-direction.R). Nothing too crazy -- it makes sure that a slendr model expressed in "forward" or "backward" slendr time units gives the same results when run via the `slim()` function in R.

5. **Here's where it gets more interesting.** [This](https://github.com/bodkan/slendr/blob/259de36d597c83bb16cda16356f93d4ab17c37d0/tests/testthat/test-time-direction.R#L121) is the R line that fails with the following error (showing a shortened slendr log written from SLiM):

```
>   slim(forward, sequence_length = 1, recombination_rate = 0)
// Initial random seed:                                                                                                                     
6632500603567566824

// RunInitializeCallbacks():

[ ... ]

// Starting run at tick <start>:
1 

tick 1 (model time 1):
  - starting the simulation
tick 1 (model time 1):
  - creating pop1(p0) with 100 individuals
tick 1 (model time 1):
  - updating map for pop1(p0)

[ ... ]

  - updating map for pop5(p4)
tick 310 (model time 310):
  - geneflow pop5(p4) -> pop4(p3) (0.005% over 40 generations)
tick 310 (model time 310):
  - updating map for pop5(p4)
ERROR (EidosInterpreter::Evaluate_MemberRef): (internal error) operand type NULL is not supported by the '.' operator.

Error on script line 497, character 29:

        source = sourceSubpop.id == GENE_FLOWS.getValue("from_id");
                             ^
Error: Unfortunately SLiM terminated before a tree sequence was saved.
See the above for an indication of where things could have gone wrong.
```

Here is the full context for the line 497 that fails -- it's a `modifyChild()` callback:

```
modifyChild() /* Place offspring within its population's boundary */ {
    // set flag marking availability for sampling
    child.tag = 0;

    if (!SPATIAL | !subpop.getValue("spatial")) return T;

    // in case of spatially non-overlapping geneflow, assign locations uniformly
    if (child.migrant) {
        // get indices of the currently active geneflow events
        active = (GENE_FLOWS.getValue("tstart_gen") <= community.tick - BURNIN_LENGTH) &
                 (community.tick - BURNIN_LENGTH <= GENE_FLOWS.getValue("tend_gen"));

        // get population of the offspring's parents
        source = sourceSubpop.id == GENE_FLOWS.getValue("from_id");    // <======== this gives the error

        // get the offspring's new population
        target = subpop.id == GENE_FLOWS.getValue("to_id");

        // if the current admixture event is non-overlapping, set the
        // childs coordinates freely
        overlap = GENE_FLOWS.getValue("overlap")[active & target & source];

        if (!overlap) {
            set_coordinates(child, subpop.id, log = F);
            return T;
        }
    }

    // make sure that offspring is generated in allowed locations on the map
    // (within it's population bounds and not in an inaccessible location)
    attempts = 0; // how many attempts have been made to place the offspring?
    do {
        attempts = attempts + 1;

        dispersal_param = subpop.getValue("dispersal");
        dispersal_fun = subpop.getValue("dispersal_fun");

        movement = dispersal_fun(dispersal_fun, dispersal_param);
        pos = parent1.spatialPosition + movement;

        found = subpop.pointInBounds(pos) & subpop.spatialMapValue("world", pos) == 0.0;
    } while (!found & attempts < MAX_ATTEMPTS);

    if (found) {
        child.setSpatialPosition(pos);
        return T;
    }

    return F;
}
```

The only purpose of this function is to:

1. For spatial models, place offspring at a coordinate not too far from its `parent1`, unless the gene flow is allowed to happen between non-overlapping populations (so its spatial location is not restricted).

2. For non-spatial models, simply accept the offspring (because it doesn't have a spatial location)

**For some strange reason, the `sourceSubpop` parameter is `NULL`.** Why is that? And why does this happen in this specific unit test and not in the many other SLiM-based unit tests in slendr? After all, this callback is called in every single on of them!

Most plausible explanations for why things broke during the SLiM 5 transition are that the SLiM script built into slendr either:

1. relies on functionality or behavior of SLiM 4.x that is no longer available in SLiM 5 (either "official" functionality or a hack that used to be possible but no longer is),

2. relies on something that changed during the SLiM 5 upgrade.

The fact that there are many other SLiM-based unit tests in slendr (including spatial tests) which pass without any issues gives me hope that the fix will be relatively simple (and hopefully will not break other things that are currently working).

I should re-read the relevant sections of the SLiM manual. The last time I wrote "pure SLiM" was... years ago when I programmed the "SLiM engine script" that's part of slendr now, and managed to forget most of what's going on. :(

So, **homework for me**:

1. In what circumstances is `sourceSubpop` NULL in `modifyChild()`?
2. Why is it NULL in this test in SLiM 5?
3. Why isn't this broken in other unit tests as well?
4. Has slendr been doing something "irregular" that only surfaces now?

bodkan:
## How to reproduce

I attach a compiled model as an archive. After unpacking it, **it can be loaded and executed in R without any additional setup like this** (this assumes that the model archive was unpacked into `~/Desktop/slim5`). In other words, until this is fixed in the compiled SLiM `script.slim` in the archive, no need to worry about other in the slendr codebase.

```
devtools::load_all("~/Projects/slendr")
init_env()

model <- read_model("~/Desktop/slim5")

# this gives the error above
slim(model, sequence_length = 1, recombination_rate = 0)
```

For further debugging, it will be probably important to bypass slendr altogether and dig right
into SLiM. **The compiled slendr model can therefore be run directly in SLiM on the command-line**:

```
slim \
  -d "MODEL_PATH='~/Desktop/slim5'" \
  -d "PATH='~/Desktop/slim5'" \
  -d SEQUENCE_LENGTH=1 \
  -d RECOMBINATION_RATE=0 \
  ~/Desktop/slim5/script.slim
```

(Opening it in SLiMgui will complain about arguments missing. Running the simulation in R like this will open the script in SLiMgui filling in the required CLI arguments accordingly `slim(model, sequence_length = 1, recombination_rate = 0, method = "gui")`.)

**Regardless of the method of execution, it gives the same error described above.**

**The same error (and no other) also pops up in all GitHub Actions runs, regardless of the setup -- Linux, macOS, Windows.**

[slim5.zip](https://github.com/user-attachments/files/20162536/slim5.zip)

bhaller:
Hi @bodkan!  I've got the script running (and erroring) on my end.  It looks to me like `sourceSubpop` is _always_ `NULL` now, inside `modifyChild()` callbacks.  Major regression there.  :-<  If your code/tests rely on it elsewhere, then yes, it is puzzling as to how anything at all works.  :->  The doc is useless on this point, sorry; clearly I need to improve it.  In a nonWF model `sourceSubpop` can perhaps be `NULL` (I'll have to check the SLiM 4.3 behavior on some edge cases where that might happen); in a WF model I don't think it ever should be.  But in SLiM 5 it _always_ is, in both WF and nonWF.  The workaround, for your purposes, is to use `parent1.subpopulation` instead of `sourceSubpop`; that is set up correctly.  I just tried it and it seems to resolve the problem.  Your script then errors later on, in `collect_metadata()`, because `SAMPLING` is `NULL`, but I assume that's nothing to do with me.  :->

So, sorry about the headaches!  I don't think I'll regard this as a problem demanding a SLiM 5.0.1 release, since there's an easy workaround and I think it will bite very few users; but I'll certainly be fixing it in the next release.  The workaround is quite harmless, though, so probably you can just adopt it and forget about it (never change it back).  :->

bhaller:
I've created a new issue at https://github.com/MesserLab/SLiM/issues/522.  It's too bad you didn't migrate `slendr` to SLiM 5 prior to its release; we could've caught this bug!  It's actually super useful to have complicated SLiM clients like `slendr` do early testing prior to a release; you exercise code paths that nobody else exercises, just because you're doing such unusual things.  :->  Perhaps we can try to co-ordinate on that in future?  I'd very much appreciate it; I hate catching bugs like this right after a release is done.  :-<

bodkan:
Hey @bhaller! Thanks for your patience (again) carefully reading my super long write up.

> Hi @bodkan!  I've got the script running (and erroring) on my end.  It looks to me like `sourceSubpop` is _always_ `NULL` now, inside `modifyChild()` callbacks.  Major regression there.  :-<

Oh no! I'm glad my weird slendr stuff helped discover another issue, I just wish it wasn't this soon after the SLiM 5 release. :( I really appreciate you finding a way to run my code (I know it's very nonstandard as far as SLiM scripts go, and is a bit of a pain to run for a non-slendr user).

> The workaround, for your purposes, is to use `parent1.subpopulation` instead of `sourceSubpop`; that is set up correctly.  I just tried it and it seems to resolve the problem.  Your script then errors later on, in `collect_metadata()`, because `SAMPLING` is `NULL`, but I assume that's nothing to do with me.  :->

Amazing, thanks for the suggestion! I'll implement it right away in this PR. And yes, the sampling thing is a problem (I think) only because I ripped out the script from the standard slendr workflow -- normally users don't run `slim -d <... stuff ...> <path to slendr/SLiM script` manually. I don't think it will show up in any other normal situation, but thank you for the heads up.

---

This is more of a side quest:

> If your code/tests rely on it elsewhere, then yes, it is puzzling as to how anything at all works. :->

So it turns out that although I do have a number of SLiM-related unit tests, they are almost all very very simple -- either single-population models, making sure slendr's weird time-unit shenanigans work as they should, doing basic tree-sequence sanity checks, comparing SLiM-vs-msprime results, that kind of thing. Even the spatial models are mostly single-population, no-gene-flow models (just to make the testing code simpler and to-the-point) which means that the problematic codepath here

```
if (child.migrant) {
... code that breaks because of the `sourceSubpop` issue
}
```

is never travelled in those tests...

... except for this one unit test featuring a complex model which is _both_ a) spatial and b) involves gene flow events, because I modified it from some tutorial vignettes. The vignettes are themselves also part of the testing suite, but I run them only after all unit tests pass -- they are much more complex, and take much longer to run. I just checked and the vignettes in question -- all spatial and all featuring gene flow events -- are also failing (again, not a surprise, given all the above).

---

> So, sorry about the headaches!  I don't think I'll regard this as a problem demanding a SLiM 5.0.1 release, since there's an easy workaround and I think it will bite very few users; but I'll certainly be fixing it in the next release. 
> [...]
> It's too bad you didn't migrate slendr to SLiM 5 prior to its release; we could've caught this bug!

No worries. And not too much of a headache, really.

But yes, too bad indeed! I feel extra bad about this one -- after all, you did suggest that I try to run slendr tests against the `multichrom` branch a few weeks ago! In my defense, I really _did_ some very surface-level tests against `multichrom`:

1. I opened slendr's `script.slim` in SLiMgui to make sure I don't get any suggestions for necessary code fixes.
2. I ran a couple of basic SLiM test scripts, which also worked OK (because they didn't travel that problematic code path).

However:

3. I got (unnecessarily) worried by the pyslim-related warnings (everything in slendr which has anything to do with SLiM 100% relies on pyslim). In the end I decided to play it safe and wait for the official pyslim release to pair it with the SLiM 5 release in slendr in one go.

:(

bodkan:
> It's actually super useful to have complicated SLiM clients like slendr do early testing prior to a release; you exercise code paths that nobody else exercises, just because you're doing such unusual things in slendr. :-> Perhaps we can try to co-ordinate on that in future? I'd very much appreciate it; I hate catching bugs like this right after a release is done. :-<

Yes, I'd like that! This is definitely that we would all benefit from. It is true that I do some very peculiar things in slendr... _(Which is sometimes why I'm feeling so hesitant and defensive about my weird SLiM code and reporting issues related to it. :D It started as a bit of a hack and it still is -- in more respects than I'm comfortable to admit -- still a hack. :D)_

But more seriously, yes. This is not the first time this kind of testing has proven useful for SLiM, so let's try to coordinate things in the future. I wonder if setting up GitHub Actions workflow dedicated to a candidate SLiM release would work. For instance, for years I've been cargo culting an absolute GHA monstrosity for automated slendr testing which currently [does this](https://github.com/bodkan/slendr/blob/rnaturalearth-fix/.github/workflows/R-CMD-check.yaml#L106-L133). The whole script runs slendr tests and building on Linux / macOS / Windows on several R version (old, current, dev branches). I guess I could add a separate GHA workflow which does this but fetches and compiles an upcoming SLiM release release? Just a thought.

bhaller:
All sounds good.  I'm glad one of your unit tests caught this bug, since it slipped through my unit tests completely!  :-O  It's really difficult to properly test software as complex as SLiM.  I wish we had a full-time QC engineer.

I like your GHA idea.  To keep it simple, how about a GHA action that just builds everything and runs slendr's standard test suite against a specified branch in SLiM, once a week say?  Right now the branch would be `master` but it would've been `multichrom` before, and pretty soon it will hopefully be `multitrait` for the next stage of things.  :->  If it's easy, it could just detect the branch where the most recent commit was made, and build/test against that branch.  That would be right 99% of the time, and if it's wrong, no worries, it'll be right a week later.  :->  Or we can just hard-code `master` for now, but then I have to remember to let you know when I switch to a side branch for a release.  Anyhow, if you know how to do that sort of thing in GHA, that'd be great!  GHA is a somewhat opaque black box to me; most of what is set up for SLiM was done by others.

And don't worry about slendr being a hack!  Real-world software almost always feel that way to the developers themselves, I think; it is a rare project indeed that has truly clean code.  SLiM has a fair bit of messiness internally, even though I do try to keep it clean; and it has a handful of hacks of breathtaking ugliness.  Such is life.  :->  Don't hesitate to ping me if you think you've found a bug – like this one!  The feedback is invaluable.  :->

bodkan:
I re-run the entire unit testing pipeline on my MacBook following your suggestion @bhaller! All tests passed! 🎉🎉🎉

```
══ Results ══════════════════════════════════════════════════════
Duration: 708.6 s

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 1125 ]
```

(BTW, I'm still amazed how fast Apple silicon is. These tests took more than 30 min on my old Intel Macbook. This is just nuts.)

That's a great sign. I will now wait what the cross-platform GHA workflows say (already running). I also included the vignettes, which are a big part of the whole unit testing pipeline and take a couple of hours to run (they contain much more complex slendr/SLiM code). If all works, SLiM5-based slendr will be good to go.

bodkan:
Yay, we got there, team!

In summary: slendr now works with the latest changes in rnaturalearth, all tests and vignettes pass with SLiM5 and Python dependencies in the slendr environment are now also updated to the latest released versions.

@bhaller: I will open a new issue about implementing a new GHA check to test against new versions of SLiM and reply there.

PR #180, opened 2025-07-13
Fix issue with ne_load() when loading small population sizes

PMassicotte:
This PR fixes an issue where `ne_load()` was incorrectly used to load small-scale data embedded in the `slendr` package:  
https://github.com/PMassicotte/slendr/blob/c4b2a12cc9b07fc5e9446880c558d478b9223391/R/interface.R#L794-L803

Previously, the code always used:

```r
map_raw <- rnaturalearth::ne_load(scale = scale, type = "land", category = "physical", destdir = ne_dir)
```

This approach failed when `scale = "small"`, because `ne_load()` now expects a GeoPackage (gpkg) file—the new format used by `rnaturalearth`—which is not compatible with the data bundled in `slendr`.

With this PR, small-scale data is now loaded directly using `read_sf()`, while `ne_download()` continues to be used for "medium" and "large" scales.

I am expecting to submit a new version of `rnaturalearth` to CRAN next week.

With regards
Phil


PMassicotte:
I have replaced the small scale internal zip file with a `gpkg` version. It will read it faster without having to unzip the file. Looks like I can not run actions on my side, but it should fix the error on Windows.

bodkan:
Awesome job, all tests passed! Thank you very much for your help here!

Issue #193, opened 2026-04-14
Python (?) / reticulate (?) / tspop (?) issues in GHA-based and Docker-based Linux runs


bodkan:
I spent the last three days debugging an extremely strange issue with failing GHA actions. This time macOS and Windows work OK, but Linux CI runs started crashing (this is almost always the other way around).

---

Briefly: setting up the slendr Python environment for GHA CI works as it always has [using code which manually replicates](https://github.com/bodkan/slendr/blob/9ebb7ca59d652f2ff0aca672421ea834766e381c/.github/workflows/R-CMD-check.yaml#L198-L209) what our [`setup_env()` function does internally](https://github.com/bodkan/slendr/blob/9ebb7ca59d652f2ff0aca672421ea834766e381c/R/interface.R#L1531-L1550). For more context: on Linux, all those Python modules are installed via `pip` by default (conda on Linux / macOS is only used to fetch a Python interpreter binary for slendr, and used to install msprime / tskit / pyslim et al. only on Windows).

However, the moment `init_env()` is called by slendr to activate that environment via reticulate / R (i.e., at any time during the CRAN build / check steps in unit test scripts or vignette rendering), slendr throws the dreaded error message stating that the Python / conda virtual environment has been found but not all Python modules are in it.

```
Python environment Python-3.12_msprime-1.4.1_tskit-1.0.2_pyslim-1.1.1_tspop-0.0.2 has
been found but it does not appear to have msprime, tskit, pyslim and tspop modules all
installed. Perhaps the environment got corrupted somehow? Running `clear_env()` and
`setup_env()` to reset the slendr's Python environment is recommended.
```

The scary thing is, I verified that those modules _are_ in that conda environment and can be, in fact, imported. This is never a good sign, and always suggests something is broken at a level of Python / conda / reticulate below slendr's feet.

---

To get a feeling for where approximately does the error happen on the R / reticulate / slendr site, I improved the error message to give much more detail. The result is even crazier -- not only is the conda environment properly activated, all msprime / tskit / pyslim / tspop are present in that environment... [_but it's only tspop which fails to be found!_](https://github.com/bodkan/slendr/blob/cf7c3d8ca633566eb5be499135010676ff491fd8/R/interface.R#L1440) (the `py_module_available()` function of reticulate tests whether a given module can be `import`-ed by the Python interpreter).

```
Python environment Python-3.12_msprime-1.4.1_tskit-1.0.2_pyslim-1.1.1_tspop-0.0.2 has
[...]

PR #194, opened 2026-04-14
Add Docker setup for slendr development and testing

bodkan:
The [recent disaster](https://github.com/bodkan/slendr/issues/193) with broken Python / reticulate / conda / GHA infrastructure finally prompted me to create a dedicated Docker container infrastructure for testing and development of slendr. I've been thinking about moving away from GHA (or maybe even GitHub itself) in the long-term and regardless of whether or not this ends up happening, having an independent self-contained infrastructure for testing slendr on Linux (which is where vast majority of high-performance simulations happens anyway) will be very useful. Just like it has been useful for the linked issue.

Furthermore, given the ever-changing landscape of a huge number of software dependencies and the constant danger of breakage due to minor (and sometimes, major) incompatibilities, I've been relying on containers for my research work for years now as a means to package entire projects in [a single artifact reproducible across all platforms](https://i.programmerhumor.io/2025/03/3d4d7cf0345843afb3e43052d37dd0ed94ac70f0c6f959d53c7df7d88c7c808b.jpeg). I've had a "using Docker -- stay tuned" in slendr docs since the very beginning and this might finally give me the infrastructure to develop the description of this approach further.

---

Examples of how I use this (utilizing the venerable, ever-present `make`):

- `make -f docker build` -- build a Docker image with all slendr dependencies
- `make -f docker bash` -- start a container with a bash shell in the current working directory (typically the slendr repo)
- `make -f docker R` -- start a container with an interactive R session console
- `make -f rstudio PORT=<number>` -- start an RStudio Server instance on the given port (typically on a remote server)
- `make -f remote-webapp SERVER=<name> PORT=<number>` -- open a "kiosk-mode" Chromium window and connect to the remote RStudio Server instance

---

For instance:

1. `make -f docker bash` starts an interactive shell
2. `make clean && make build && make check` in that shell executes the CRAN R CMD build / check workflow

Or:

1. `make -f docker R` starts an interactive R console
2. `devtools::load_all()` in the R console loads the development version of slendr for a normal R package development workflow (good enough for quick and simple testing)

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -25,4 +25,6 @@
 .codecov.yml
 ubuntu_deps.txt
 ^vignettes/vignette-09-paper\.Rmd$
-TAGS
+Dockerfile
+docker
+tags
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,106 @@
+FROM rocker/rstudio:4.5.3
+
+LABEL maintainer="Martin Petr <mp@bodkan.net>"
+
+############################################################
+# setup the base system
+############################################################
+
+ENV DEBIAN_FRONTEND="noninteractive"
+
+RUN apt-get update -y \
+    && apt-get upgrade -y \
+    && apt-get install -yqq --no-install-recommends \
+        build-essential \
+        cmake \
+        curl \
+        gdal-bin \
+        git \
+        htop \
+        iputils-ping \
+        less \
+        libblas-dev \
+        libbz2-dev \
+        libcurl4-openssl-dev \
+        libffi-dev \
+        libfftw3-dev \
+        libfontconfig1-dev \
+        libfribidi-dev \
+        libgdal-dev \
+        libgit2-dev \
+        libglpk-dev \
+        libglu1 \
+        libgsl-dev \
+        liblapack-dev \
+        libharfbuzz-dev \
+        libmagick++-dev \
+        libnlopt-cxx-dev \
+        libpopt-dev \
+        libreadline-dev \
+        librsvg2-dev \
+        libsnappy1v5 \
+        libssl-dev \
+        libsqlite3-dev \
+        libtiff5-dev \
+        libudunits2-dev \
+        libunwind-dev \
+        libwebp-dev \
+        libxml2-dev \
+        libxt6 \
+        libzmq5 \
+        libzstd-dev \
+        man-db \
+        parallel \
+        rename \
+        tmux \
+        tree \
+        unminimize \
+        vim \
+        wget \
+        zlib1g-dev
+
+# fix 'Errors were encountered while processing: fontconfig' during unminimize
+RUN fc-cache -f; yes | unminimize
+
+# the container is intended as a dedicated environment to be run in a rootless setting
+ENV HOME="/root"
+
+############################################################
+# compile and install third-party software dependencies
+############################################################
+
+# compile SLiM
+RUN cd /tmp; wget https://github.com/MesserLab/SLiM/archive/refs/tags/v5.1.tar.gz -O slim.tar.gz; \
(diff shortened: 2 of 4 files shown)


bhaller:
I'm always impressed by your competence in this arena.  I have no idea how you do all this; with this kind of thing I just immediately glaze over.  This sounds very useful.  (I'm more skeptical of the idea of moving off of GitHub, though; what would be the argument for that?)

bodkan:
> I'm always impressed by your competence in this arena. I have no idea how you do all this; with this kind of thing I just immediately glaze over. This sounds very useful.

Thank you for the compliment! :) I wish this wasn't driven primarily by a) anxiety driven by having to juggle what appears to be an ever-shifting changes in various software dependencies and b) my obsessive compulsion to pursue and maintain stability and order. Both a) and b) make for a horrific combination, let me tell you! 😰

That said, even outside of that personal dimension, Docker really has proven to be very useful, especially in my research environments in which I have no control about what happens to our compute servers (random IT system upgrades, missing header files, etc.). Once I took a Docker container I built for a project in one institution's servers (together with some completely unportable ancient legacy code), moved it to another institution's servers, and when I saw that everything worked without any change and without having to install anything, there was no going back.

It can be also very useful for software development and testing like I described above (though it's a Linux-specific technology) but honestly, it's much more valuable as a research tool. For instance, in the context of SLiM development, I don't think you're missing very much!

> (I'm more skeptical of the idea of moving off of GitHub, though; what would be the argument for that?)

I worry it would make me go too much into the philosophical and -- even more dangerously! -- into the political. 😅 And even then, I'm not sure if I would be able to put together a coherent answer. Maybe we can have a chat about that when you're back in Copenhagen for SMBE. Assuming the world is still spinning by then. :)

bhaller:
> > I'm always impressed by your competence in this arena. I have no idea how you do all this; with this kind of thing I just immediately glaze over. This sounds very useful.
> 
> Thank you for the compliment! :) I wish this wasn't driven primarily by a) anxiety driven by having to juggle what appears to be an ever-shifting changes in various software dependencies and b) my obsessive compulsion to pursue and maintain stability and order. Both a) and b) make for a horrific combination, let me tell you! 😰

:-O  Well, everything about one's nature can be both a blessing and a curse.  :->

> That said, even outside of that personal dimension, Docker really has proven to be very useful, especially in my research environments in which I have no control about what happens to our compute servers (random IT system upgrades, missing header files, etc.). Once I took a Docker container I built for a project in one institution's servers (together with some completely unportable ancient legacy code), moved it to another institution's servers, and when I saw that everything worked without any change and without having to install anything, there was no going back.

Yeah, that's amazing.  I've never needed it, so I've never learned it, but it seems very powerful.  I tend to strongly resist learning things until I need them.  I'm a big believer in specialization; productivity is much higher if you stay in your own lane, do what you're good at, and don't get distracted by shiny things elsewhere unless you need them.  That's why I've never learned Python; I just wouldn't use it often enough for the time investment to pay off, I think.  But sometimes it is limiting, for sure.

> It can be also very useful for software development and testing like I described above (though it's a Linux-specific technology) but honestly, it's much more valuable as a research tool. For instance, in the context of SLiM development, I don't think you're missing very much!

Yeah, exactly.  :->

> > (I'm more skeptical of the idea of moving off of GitHub, though; what would be the argument for that?)
> 
> I worry it would make me go too much into the philosophical and -- even more dangerously! -- into the political. 😅 And even then, I'm not sure if I would be able to put together a coherent answer. Maybe we can have a chat about that when you're back in Copenhagen for SMBE. Assuming the world is still spinning by then. :)

Fair enough.  I do hope we'll have time for a bit more birding, and a lot more hanging out, when I'm in Copenhagen.  :->

PR #195, opened 2026-04-16
Add uv as a fallback for systems with a broken conda / Python / reticulate interaction

bodkan:
_(Ignore the name of the git branch. GitHub doesn't allow changing branch names for opened PRs. We're not "switching" anything.)_

This is an attempt to provide a workaround for a broken interaction between the Python interpreters compiled by the conda project, the way its linked to system libraries (as described [here](https://discuss.python.org/t/moving-expat-version-or-symbol-checks-from-buildtime-to-runtime/105769) in detail) and the reticulate package which embeds a Python interpreter into an R session -- occasionally encountered on a subset of Linux systems.

The entire story so far is described [here](https://github.com/bodkan/slendr/issues/193), and is also related to an [earlier issue](https://github.com/bodkan/slendr/blob/289e2c6fb850c4b9f861e3d9d32c7c03a16978db/NEWS.md?plain=1#L28-L35) discovered in this context.

Because I don't see a way to solves this at the low-level where the problem actually happens (in fact, [the python.org bugtracker discussion](https://discuss.python.org/t/moving-expat-version-or-symbol-checks-from-buildtime-to-runtime/105769) makes it hard to even see how this could be fixed at the level of slendr), I decided to implement a (semi-)automatic fallback mechanism.

### Updated Python activation setup

1. For systems where no issue occurs (so far, all macOS tests, all Windows tests, most Linux tests), nothing changes. The following combo works just as before:

```
library(slendr)
setup_env()
init_env()
```

4. For systems where the reticulate-based embedding of the Python interpreter breaks (i.e., working `setup_env()` creation of a Python environment, broken embedding of Python into R in `init_env()`), users can continue using slendr's tskit/pyslim/msprime/tspop based features by doing the following:

```
library(slendr)
init_env(uv = TRUE)
```

Note that this skips the creation of a dedicated Python environment `setup_env()` entirely (because in situations where the embedding of Python into R breaks, creating the environment doesn't make sense). Instead, `init_env()` now contains a new argument `uv =` which, when set `TRUE`, uses the [relatively new functionality of the reticulate package to create ephemeral, throwaway Python environments](https://posit.co/blog/reticulate-1-41/). Because those environments do not utilize Python interpreter binary distributed by conda (what causes our issues right now), all scripts work without an issue.

Importantly, this change is completely backwards compatible, as calling `init_env()` (defaulting to `init_env(uv = FALSE)`) utilizes the reticulate-based Python machinery we've been using for years now.

Additionally, because I find the `init_env(uv = TRUE)` a bit annoying and easy to forget, there's also the possibility to set the environment variable `SLENDR_UV = TRUE` (for instance, in the standard `~/.Renviron` file), which makes `init_env()` (without arguments) behave in the exact same way as `init_env(uv = TRUE)`. This could be useful in containers, HPC settings which run into the same Python/conda/reticulate problems, etc.

### Further comments on uv, long-term perspective on integrating Python, etc.

The RStudio team behind reticulate has been pushing quite a bit for the adoption of uv as a new and popular tool to manage Python installations and environments. In fact, they seem to be promoting the usage of ephemeral throwaway Python environments for every individual script, utilizing the fact that uv is extremely fast and uses caching (so downloading and installing Python for every script run is quite fast). Relying on fixed, reused, more stable Python environments (which has been our strategy since the very beginning) seems to be a bit discouraged.

As a matter of fact, at first, I briefly considered porting the Python infrastructure of slendr to uv entirely. I decided not to do that for a number of reasons:

1. The primary case of using conda to fetch Python interpreter and install packages such as msprime or tskit is to allow slendr usage in challenging computational environments, namely HPC systems or Windows systems where users don't have to ability to install header libraries such as GSL, etc. This is the primary reason why conda was created in the first place, and uv doesn't help with that. uv is fast, but doesn't solve the distribution problem. (Side note: I'm a bit puzzled that this is never brought up in discussion about uv?) This alone was a reason why I originally disregarded uv for slendr completely -- its primary benefits are not that useful for us, the drawbacks potentially substantial.

2. Even though ephemeral uv-based Python environments can be downloaded and created very fast, an important application for slendr is simulation-based-inference. Although uv caching helps a lot with performance, sharing a single stable Python installation across thousands or even millions of simulation runs seems much more sense.

3. Although I used to use uv outside of R for a long time, I couldn't help to worry about the long-term stability of the tool given the VC backing of Astral (the developers of uv). Indeed, the recent [acquisition of Astral by ClosedAI](https://openai.com/index/openai-to-acquire-astral/) makes me even more skeptical about the long-term fate for uv.

In summary, uv provides a very useful emergency fallback for Linux (although still only for systems which do not suffer from issues related to installing libraries in HPC environments where users do not have root privileges), but it seems like a good idea not to switch to it completely.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -161,9 +161,9 @@ jobs:
               "\n",
               "RSTUDIO_PANDOC=", normalizePath(dirname(pandoc::pandoc_bin()), winslash = "/"),
               "\n",
-              "GTIFF_SRS_SOURCE=EPSG # silence GDAL warning"
+              "GTIFF_SRS_SOURCE=EPSG # silence GDAL warning",
               "\n",
-              "SLENDR_UV=TRUE
+              "SLENDR_UV=TRUE"
             ),
             con = paste0(normalizePath(path.expand("~"), winslash = "/"), "/.Renviron"))
         shell: Rscript {0}
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -117,11 +117,12 @@ jobs:
           cmake .
           make slim
 
-      - name: Populate PATH with compiled binaries on Unix
+      - name: Set necessary variables in ~/.Renviron on Unix
         if: runner.os != 'Windows'
         run: |
           export PATH="$HOME/deps/SLiM:$HOME/deps/AdmixTools-7.0.2/bin:$PATH"
           echo "PATH=$PATH" >> ~/.Renviron
+          echo "SLENDR_UV=TRUE" >> ~/.Renviron
 
       - name: Install SLiM on Windows
         if: runner.os == 'Windows'
@@ -151,7 +152,7 @@ jobs:
       #  run: dir ${{ runner.temp }}\msys64\usr\bin
       #  shell: cmd
 
-      - name: Populate PATH with compiled binaries on Windows
+      - name: Set necessary variables in ~/.Renviron on Windows
         if: runner.os == 'Windows'
         run: |
           writeLines(
@@ -161,6 +162,8 @@ jobs:
               "RSTUDIO_PANDOC=", normalizePath(dirname(pandoc::pandoc_bin()), winslash = "/"),
               "\n",
               "GTIFF_SRS_SOURCE=EPSG # silence GDAL warning"
+              "\n",
+              "SLENDR_UV=TRUE
             ),
             con = paste0(normalizePath(path.expand("~"), winslash = "/"), "/.Renviron"))
         shell: Rscript {0}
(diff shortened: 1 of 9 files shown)


bodkan:
OK. I declare victory. Here's the current status of things:

1. All GitHub Actions runs on Linux and macOS now utilize the new uv-based ephemeral Python environments.
2. As expected, Windows GHA runs with uv run into problems, but not because R or slendr -- the runs crash with "PowerShell not found" during reticulate's uv installation procedure. This isn't worth debugging for me -- for Windows GHA, I resort to the original (and still default) workflow based on a stable, non-ephemeral Python environments.

Importantly, local R CRAN checks -- my own Mac, a Linux Docker container running on my Mac, a dedicated Linux server, and a Windows instance running in a virtual machine on my Mac -- all work with the new uv-based ephemeral environments.

Given that the primary goal of this PR was to fix occasional crashes on some very specific Linux systems, I consider this a success.

---

For completeness, the conda/Python dedicated environments (stable and persistent, i.e. non-ephemeral) are still the default, and calling `init_env()` as always will always utilize those environments (first created by our `setup_env()` function). The new uv-based engine is utilized upon calling `init_env(uv = TRUE)` or by calling `init_env()` with the environment variable `SLENDR_UV` set to `"TRUE"` (typically in something like the standard `~/.Renviron` file).

---

I will let it sit for a while and include these changes in the next slendr release. This whole thing was an emergency fix which delayed some research projects I had planned for this week, so getting back to those next week will be a good opportunity to test all of this in practice, not just in the standardized R CRAN check pipelines.

PR #197, opened 2026-04-22
Store information about dependency versions 'externally' under `inst/deps`

bodkan:
It has become really bothersome having to trace and update dependency versions of Python modules and SLiM throughout `R/zzz.R` (always cracks me up that this is an "officially" named file for R packages) and elsewhere. Let's try moving all of this information under `inst/deps`, reading everything upon loading slendr with every `library(slendr)`, and storing all versioning information in a single global list for use throughout the entire codebase.

---

TODO: Can I use this for GitHub Actions pipelines too?

EDIT: I guess I could use this: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts, utilizing the `$GITHUB_WORKSPACE` as an absolute prefix to `inst/deps`?

image


bodkan:
SLiM v5.2 was just released, so I updated all unit tests across all three platforms (wherever outputs differed because of the random number generation changes).

